PR #25, opened 2019-12-05
Hygen setup

clarknoah:
Added Hygen Templates


browser/_templates/generator/help/index.ejs.t
```diff
@@ -0,0 +1,5 @@
+---
```
ZakkMan: Do we need these? What purpose do they serve?

browser/_templates/react-boilerplate/new/index.ejs.t
```diff
@@ -0,0 +1,19 @@
...
+import React from 'react';
+import ReactDOM from 'react-dom';
+import './index.css';
+import App from './App';
+import {
+  BrowserRouter as Router
+} from "react-router-dom"
+import * as serviceWorker from './serviceWorker';
+
+ReactDOM.render(<Router><App /></Router>, document.getElementById('root'));
```
ZakkMan: Setting up React Router should be a separate PR

browser/_templates/react-boilerplate/new/app.ejs.t
```diff
@@ -0,0 +1,26 @@
+---
```
ZakkMan: Do we need this? there's already an `App.js`

browser/_templates/component-const/new/reactComponent.ejs.t
```diff
@@ -0,0 +1,22 @@
+---
+to: src/components/<%= name %>/<%= name %>.js
+---
+
+import React from 'react';
+import "./<%= name %>.css";
+
+// Function based React Component
+const <%= name %> = (props) => {
```
ZakkMan: Is there a way to uppercase the component names? i.e. name the component `Button` if I run `hygen component-const new button`.

browser/_templates/react-boilerplate/new/index.ejs.t
ZakkMan: Oh wait - I see what you're doing!

We're going to setup index, app, etc in a later PR - we only need a generator for repeated actions (like creating components) - good thinking though!

clarknoah: Oh wow, this is embarrassing, the entire react-boilerplate was accidentally included from some other project I was working on, totally didn't even see that I left it in the directory!

browser/_templates/component-const/new/reactComponent.ejs.t
clarknoah: When you use Hygen (as the script is currently configured), in order to upper case the component names, you must do it when you write it into command line - EG: 
`hygen component-const new --name Button`
--> * src/Component/Button/Button.js*

I had intially set it up where it would capitalize, but if the variable name is more than one word, it will only capitalize the first character in the string,  EG: 
`hygen component-const new --name newbutton`
--> * src/Component/Newbutton/Newbutton.js*
so I decided against using it. Would you like me to change it?

ZakkMan:
@ZakkMan

ZakkMan:
@clarknoah

Issue #7, opened 2019-12-04
Draft a readme for the client


ZakkMan:
We currently just have the default readme from create-react-app. We need one that includes:

* Description/background
* Setup and installation instructions
* Background on the component library
* Documentation on the Hygen setup (#3)

wchrapcynski:
@ZakkMan There is currently quite a bit in the readme already that I was thinking that I needed to do. Until we have the application built, I'm not sure what more I can add right now.

ZakkMan:
@wchrapcynski Theres a first draft of the readme for the whole project - You can move some of that into the readme for the client if you think that's where it makes more sense. I think the biggest contribution here is instructions for everyone on getting this app up and running on their laptops. This documentation will change as the app gets built out. But we also need instructions on how to get it set up as it it

wchrapcynski:
@ZakkMan Are you still changing the folder for the client from browser to client? You mentioned it yesterday. Since it's still browser, I'll call it that for now in the readme.

clarknoah:
Bill, below this line is the documentation for hygen:
# Hygen Setup 
To use Hygen with your project, you need to first install it globally using NPM:

> npm install hygen -g 

Once you have installed hygen, you will be able to utilize it within the `client` directory. 

# Hygen Generators 
There are two generators available for hygen, the `component-class` generator, and the `component-const` generator

## `component-class` 
The `component-class` generator creates four files within the src/components/ directory and automatically builds for you the component folder, as well as the four files you need to get started on a component. 

This generator creates a React `class` component, with a few lifecycle boilerplate functions built-in. 

To run the generator for this component, run the following command FROM the `client` directory (it will be the directory that contains the _template folder within it):
> hygen component-class new --name NameOfComponent

The `component-class` generator is used to create the following files: 
###  src/components/NameOfComponent/NameOfComponent.css
Creates CSS Boilerplate
```css
.NameOfComponent {

}
```
###  src/components/NameOfComponent/NameOfComponent.js
React Component
```javascript
import React, {Component} from 'react';
import "./NameOfComponent.css";


// Class Based React Component
class NameOfComponent extends Component{
  constructor(props){
    super(props);
    console.log(props);

    // Default CSS class to apply to the Component
    this.state = {
      classList: "NameOfComponent"
    };
  }


  // Runs after Component is loaded in the broswer
  componentDidMount(){}


  // Runs after a component has been updated
  componentDidUpdate(){}


[...]

Issue #6, opened 2019-12-04
Draft a readme for the API


ZakkMan:
We need some documentation for the backend API. It should include:

* Description/background
* List of technologies used
* How to setup the API (including database setup)

PR #34, opened 2019-12-05
Created editorconfig

TylerEikenberg:
Created editorconfig file in root of directory. Added rules to handle JS syntax and python syntax, as well as markdown. I'm not totally positive but everyone on the project may need to have the vscode editorconfig extension installed? I'll look into this.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,25 @@
+# http://editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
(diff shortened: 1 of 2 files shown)


ZakkMan:
@TylerEikenberg Thanks! Let me know if we need to have that plugin installed - if so, can you update the documentation, working with @wchrapcynski and @MobolanleAdebesin to update #7 and #6?

TylerEikenberg:
@ZakkMan Will do, researching now. Thanks for the feedback :+1:

PR #36, opened 2019-12-06
Documentation - Client Readme (first draft) PR

wchrapcynski:
Here's the first draft of the client readme for issue #7. I think I have the basics in here ready to go. Let me know what you think.


wchrapcynski:
> @wchrapcynski can you call the file just `README.md` and make sure it's in the `client/` directory?
> 
> Also, it looks like you have some changes to the `Pipfile` - can you get ride of those?

I see that you already did this as well as change the name of the readme file? Also, the browser folder still seems to be there.

Issue #2, opened 2019-12-04
Component: Pill Issue


ZakkMan:
![Screen Shot 2019-12-04 at 9 27 50 AM](https://user-images.githubusercontent.com/4991553/70150703-6b482c00-1678-11ea-9005-9b663bc3399b.png)
<img width="232" alt="Screen Shot 2019-12-06 at 2 46 58 PM" src="https://user-images.githubusercontent.com/55462909/70351639-61b8f280-1837-11ea-9d54-c1757661704c.png">
<img width="252" alt="Screen Shot 2019-12-06 at 2 48 14 PM" src="https://user-images.githubusercontent.com/55462909/70351679-76958600-1837-11ea-934d-c8f2272dfd38.png">

Pill Component.

Properties:
* `text`: string
* `active`: boolean
* `onStatusChange`: callback
* `iconPath`: String
* `iconActive`: boolean
* `iconOffset`: boolean

If `active` is set to true, the text color should be white and the background should be pink. If `active` is set to false, the text should be black and the background should be white.

The `onStatusChange` callback should be called after the status (i.e `active`) property is changed.

If `iconActive` is `true` then an icon should displayed on the right side, using `iconPath`, like the second image. If `iconOffset` is true then the icon should be offset towards the top of the pill like the third image. 

This should have a hover effect so that the background turns a very light pink when the user hovers over it.

joshseven:
Duplicate component that will be covered in component buttons

PR #41, opened 2019-12-06
Pill Component PR

ZakkMan:
I built out the Pill component per #2.

During the build, I had to deviate from the spec like this ... for this reason ...

I've tested it on mobile and it works just fine.

I had some questions about the best way to handle X, here is what I thought of:

* One option
* Another option
* a third option

### DIFF
--- a/browser/src/components/Pill/Pill.js
+++ b/browser/src/components/Pill/Pill.js
@@ -1,0 +1,9 @@
+import React, { Component } from "react";
+
+class Pill extends Component {
+  render() {
+    return <div>this is a Pill component</div>;
+  }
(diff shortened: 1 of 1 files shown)


browser/src/components/Pill/Pill.js
```diff
@@ -0,0 +1,9 @@
+import React, { Component } from "react";
+
+class Pill extends Component {
+  render() {
+    return <div>this is a Pill component</div>;
+  }
+}
```
ZakkMan: Hey, this component doesn't appear to have or need any state - could we turn it into a function component?

Issue #45, opened 2019-12-06
Gather image assets Issue


wchrapcynski:
Exporting all of the images from the mock for building Storybook components.

wchrapcynski:
It looks like there were some missing images that are in App.css but not the assets folder. I have those committed now and will set up a new pull request.

PR #47, opened 2019-12-06
Asset Gathering PR

wchrapcynski:
Per issue #45, these are asset images gathered for building components. I've put them all in an assets/images folder in the root. I'm not sure if this is the right place for it but I wanted to make sure these were available to the group at large.


ZakkMan:
Thanks Bill - this looks good!

Issue #101, opened 2019-12-06
Create CSS variables Issue


wchrapcynski:
Create CSS variables for App.css that will be defaults for the entire project. 
Here's an example for using one:

background-color: var(--yellow);

Adding in variables for things that do not change and will be used across all components.

PR #103, opened 2019-12-06
Css variables PR

wchrapcynski:
Added CSS variables for a default color palette, images, and font weights in App.css all for issue #101 . More will probably be needed but I think this is a good start for everyone to use.


client/src/App.css
```diff
@@ -1,3 +1,42 @@
...
+    --social-facebook: url("../../assets/images/social_facebook.png");
+    --social-instagram: url("../../assets/images/social_instagram.png");
+    --social-pinterest: url("../../assets/images/social_pinterest.png");
+    --slash-background: url("../../assets/images/splash_background.png");
+    --text-box-search: url("../../assets/images/text_box_search.png");
+    /* Font Weights */
+    --paragraph-weight: 14px;
+    --banner-weight: 80px;
+    --link-weight: 24px;
+    --title-weight: 36px;
```
ZakkMan: These need to be in units of 100 (i.e. 100, 200, 300, etc up to 900) - like we talked about in person

wchrapcynski: Okay. This is fixed and should be ready. There's a pull req for it currently.

wchrapcynski: @ZakkMan Can we get this merged ASAP so that everyone can get to using them? Thanks!

ZakkMan:
@wchrapcynski Great Work!

Issue #15, opened 2019-12-05
Component: Logo Issue


tissakhosla:
![image](https://user-images.githubusercontent.com/53238865/70267122-c9a20700-176b-11ea-9860-a1e2716b2ad4.png)

This component just displays the logo in various locations. It should not be clickable by default.

Properties:
* `clickable`: boolean
* `onStatusChange`: callback

If clickable is true, callback will go to the root of the website.

PR #106, opened 2019-12-06
Component: Logo PR

torykling:
#15 

I built a logo component with a size property to make it smaller for nav and a clickable property to add a link to return to the home page. 

Question: Do the sizes seem right? I set the default size to be 20vh by 20vh and the small to be 10vh by 10vh.


client/src/components/Logo/Logo.js
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+import { Link } from "react-router-dom";
+import "./Logo.css";
+
+// Function based React Component
+const Logo = props => {
+  // Default Class to apply to Component
+  let classList = `Logo`;
+  if (props.small) {
+    classList += ` smallLogo`;
```
ZakkMan: Can you change this class name to `small-logo`? We generally want our classes to be kebab case

client/src/components/Logo/Logo.js
```diff
@@ -0,0 +1,23 @@
...
+  }
+  if (props.clickable) {
+    return (
+      <div className={classList}>
+        <Link to="/" />
+      </div>
+    );
+  } else {
+    return <div className={classList}></div>;
+  }
```
ZakkMan: This is awesome! Can you make it so someone can pass in the destination URL as a prop, with the default value as `"/"`? We'll almost always want the logo to link to the homepage (if it is clickable) but there may be cases where we want to change that.

ZakkMan: Let's call the prop `to`, to keep it consistent with React Router

client/src/components/Logo/Logo.css
```diff
@@ -0,0 +1,13 @@
+.Logo {
+  background-image: url("./logo.png");
+  background-size: cover;
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 20vh;
+  height: 20vh;
```
ZakkMan: I would do this in pixels or rems - maybe like `3rem` or `4rem` for large and `2rem` for small?

torykling:
@ZakkMan Thanks for reviewing this! I made some changes. Do you think these are okay? 

Also am I supposed to fill out a new pull request somehow to resubmit for review or should I just write a comment like this?

Issue #64, opened 2019-12-06
Component: Topic Pill Issue


cjohnston96:
<img width="183" alt="Screen Shot 2019-12-06 at 2 07 10 PM" src="https://user-images.githubusercontent.com/30026036/70348963-ca9d6c00-1831-11ea-893b-132f47991dd0.png">

Text inside of div possibly as **props**

nickmckenney:
Working on Pill Atomic Component
 per #64

nickmckenney:
This issue seems fixed. Worked on changing the props to pass down background color as well as text instead of using css

Issue #68, opened 2019-12-06
Component: Pill Select Component issue


clarknoah:
<img width="580" alt="Screen Shot 2019-12-06 at 2 13 40 PM" src="https://user-images.githubusercontent.com/1622159/70349431-b6a63a00-1832-11ea-9bb0-99a11c9e66cb.png">

- Features
    - Component should be clickable. When the component is clicked, component background color should change
- Inputs
    - Component should receive an input property of “interest={interestName}”
- Internal Actions
    - When user clicks on component, component should change colors
    - Clicking should change the state of the component boolean value with updateState=()=> . 
    - {this.setState({selected:!this.state.selected})
updateState should change the CSS class assigned to it as well
- Outputs
    - Component should have output property “selected={parentComponentFunction}”, which will return an object {key: ComponentKey, selected: Boolean}

PR #107, opened 2019-12-06
Component:Topic Pill PR

nickmckenney:
per #64 

Screen Shot 2019-12-06 at 5 03 07 PM
Worked on a responsive pill component. Made text, height, width responsive according to the viewport. 
Questions: Should this pill have a click event in a storybook that changes it color?


client/src/components/Pill/Pill.js
```diff
@@ -0,0 +1,16 @@
...
+import "./Pill.css";
+
+// Function based React Component
+const Pill = props => {
+  // Default Class to apply to Component
+  let classList = `PillContainer`;
+
+  return (
+    <div className={classList}>
+      <h1 className="pillContainer">Topic</h1>
```
ZakkMan: Can you make the CSS class `pill-container` (CSS Class names should be lowercase and kebab case)

client/src/components/PillTopic/Pill.test.js
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import { shallow, configure } from "enzyme";
+import PillTopic from "./PillTopic.js.js";
```
ZakkMan: Here it's called `PillTopic`, but the file is called `Pill` and the component is called `Pill`. Can you make sure this works?

client/src/components/Pill/Pill.stories.js
```diff
@@ -0,0 +1,5 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import Pill from "./Pill";
+
+storiesOf("Pill", module).add("Default", () => <Pill />);
```
ZakkMan: We need a way to pass down the color and the text as a prop. Can you implement that?

client/src/components/Pill/Pill.css
```diff
@@ -0,0 +1,16 @@
+.pill-container {
+  height: 1.061vw;
+  width: 7.071vw;
```
ZakkMan: Can you use `px` or `rem` for the sizing?

client/src/components/Pill/Pill.css
```diff
@@ -0,0 +1,16 @@
...
+  position: absolute;
+  left: 0%;
+  right: 0%;
+  top: 0%;
```
ZakkMan: Why is this absolutely positioned like this? It doesn't need to be

client/src/components/Pill/Pill.css
```diff
@@ -0,0 +1,16 @@
...
+  font-family: non-serif;
+  bottom: 0%;
+  font-size: 0.707vw;
+  font-family: sans-serif;
+  font-weight: 350;
```
ZakkMan: For the font family and font weight, can you use one of the CSS variables?

client/src/components/PillTopic/Pill.test.js
```diff
@@ -0,0 +1,16 @@
...
+// We will describe a block of tests
+describe("PillTopic component", () => {
+  // we will write one individual test
+  it("should render as expected", () => {
+    // Shallow rendering renders a component without rendering any of its children
+    const component = shallow(<PillTopic />);
+    // We create an assertion within the test that checks if our component renders our name prop
+    // expect(component.contains('Your name')).toBe(true)
+  });
+});
```
ZakkMan: This looks like an extra file and folder - can we delete these?

client/src/components/Pill/Pill.css
```diff
@@ -0,0 +1,16 @@
+.pill-container {
+  height: 1.061vw;
+  width: 7.071vw;
+  position: absolute;
+  left: 0%;
+  right: 0%;
+  top: 0%;
+  font-family: non-serif;
+  bottom: 0%;
+  font-size: 0.707vw;
```
ZakkMan: Can you use `px` or `rem` for the font size?

client/src/components/Pill/Pill.css
ZakkMan: Also font weight has to be a unit of 100 (i.e. 100, 200, 300, 400, 500, all the way up to 900)

client/src/components/Pill/Pill.css
```diff
@@ -0,0 +1,16 @@
...
+  font-family: sans-serif;
+  font-weight: 350;
+  text-align: center;
+  color: white;
```
wchrapcynski: Make sure that you are using the CSS variables when referring to colors. 
So, here it would be color: var(--white); 
The color isn't actually pure white.

nickmckenney:
worked with Noah on the current issues and resolved everything on the issue per #68

Issue #84, opened 2019-12-06
Component: Carousel Dot Component Issue


clarknoah:
<img width="289" alt="Screen Shot 2019-12-06 at 2 37 29 PM" src="https://user-images.githubusercontent.com/1622159/70351048-0b977f80-1836-11ea-8670-de5dc559e1c8.png">

<img width="572" alt="Screen Shot 2019-12-06 at 2 38 11 PM" src="https://user-images.githubusercontent.com/1622159/70351057-13572400-1836-11ea-8816-ad23328bb051.png">

- Features
    - Component should have one dot highlighted and the others white
- Inputs
    - `numberOfDots`: number of inputDots
    - `currentLocation`: currentlyHighlightedPosition
    - `highlightedDotColor`: color for highlighted dot 
    - `notHighlightedDotColor`: color for circles not highlighted
    - `hasBorder`: boolean which will determine whether to have a border for the circle if unselected
- Internal Actions
    - On `componentDidUpdate()`
    - Check currentLocation prop and update pill accordingly
- Outputs
    - `onStatusChange`: callback

shelbyvjacobs:
Also see **Issue #11** - Component: Navigation Dots

caioingber:
This has been complete per notes above and closed PR #111

PR #111, opened 2019-12-07
#84 CarouselDots component completed PR

clarknoah:
CarouselDots component has been completed!


ZakkMan:
@clarknoah Make sure you assign yourself the PR

client/src/components/CarouselDots/CarouselDots.css
```diff
@@ -0,0 +1,18 @@
...
+.CarouselDot {
+}
+
+.UnselectedDot {
+
+}
+
+.SelectedDot {
+
+}
```
ZakkMan: If we don't need these classes, then lets delete them!

client/src/components/CarouselDots/CarouselDots.js
```diff
@@ -0,0 +1,124 @@
...
+
+    // Default CSS class to apply to the Component
+    this.state = {
+      classList: "CarouselDots",
+      numberOfDots: props.numberOfDots,
+      activeLocation: props.activeLocation,
+      activeDotColor: props.activeDotColor,
+      inactiveDotColor: props.inactiveDotColor,
+      hasBorder: props.hasBorder,
+      borderWeight: props.borderWeight,
```
ZakkMan: If this data is in `props`, then there's no reason to assign it to `state` unless it's data that is going to change. It doesn't look like any of this data is changing

client/src/components/CarouselDots/CarouselDots.js
```diff
@@ -0,0 +1,124 @@
...
+
+
+  // Runs after Component is loaded in the broswer
+  componentDidMount() {
+    let height = ReactDOM.findDOMNode(this).parentNode.offsetHeight;
+    this.setState({
+      height: height,
+      width: height
+    })
+  }
```
ZakkMan: It's unclear to me why we need this and it seems a little dirty. What is this doing and is there another way to achieve this?

For instance, the height and width can probably just be fixed in the CSS

client/src/components/CarouselDots/CarouselDots.js
```diff
@@ -0,0 +1,124 @@
...
+    //console.log(this.state.dots)
+    // console.log(props);
+    // let activeLocation = this.props.activeLocation;
+    // let previousLocation = this.state.activeLocation;
+    // let newLocation = activeLocation !== previousLocation;
+    // console.log(activeLocation, previousLocation, newLocation);
+    // if(newLocation){
+    //   this.setState({activeLocation});
+    // }
+  }
```
ZakkMan: Let's delete this if it's not doing anything.

client/src/components/CarouselDots/CarouselDots.js
```diff
@@ -0,0 +1,124 @@
...
+  generateDots=()=>{
+    let numberOfDots = this.state.numberOfDots;
+    let activeDot = this.state.activeLocation;
+    let dotArray = [];
+    let height = this.state.height;
+    let width = this.state.width;
+    let activeFill = this.state.activeDotColor;
+    let inactiveFill = this.state.inactiveDotColor;
+    let hasBorder = this.state.hasBorder;
+    let borderWeight = this.state.borderWeight;
```
ZakkMan: This is a lot of stuff coming from `state`/`props` - can we clean this up with destructing?

```js
let { a, b, c } = letters
```

client/src/components/CarouselDots/CarouselDots.js
```diff
@@ -0,0 +1,124 @@
...
+            <svg key={i} height={height} width={width}>
+              <circle onClick={()=>{this.dotClicked(i)}} cx={height/2} cy={width/2} r={width/5}  fill={inactiveFill} />
+            </svg>
+          dotArray.push(dot);
+        }
+      }
+
+    }
+    return dotArray.map(val=>val);
+  }
```
ZakkMan: There's a lot of repeated code here - can you find a way to achieve this without so many `if`/`else` statements?

client/src/components/CarouselDots/CarouselDots.js
```diff
@@ -0,0 +1,124 @@
...
+    })
+  }
+
+  // Runs right before a component is removed from the DOM
+  componentWillUnmount(){}
+
+  generateDots=()=>{
+    let numberOfDots = this.state.numberOfDots;
+    let activeDot = this.state.activeLocation;
+    let dotArray = [];
```
ZakkMan: Can we call this variable `dots` instead of `dotArray`?

client/src/components/CarouselDots/CarouselDots.js
ZakkMan: You can achieve this without any `if`/`else` statements and just maybe 2 ternary statements

client/src/components/CarouselDots/CarouselDots.js
```diff
@@ -0,0 +1,91 @@
+import React, { Component } from "react";
+import "./CarouselDots.css";
+import ReactDOM from "react-dom";
+
+class CarouselDots extends Component {
+  constructor(props) {
+    super(props);
+    console.log(props);
```
ZakkMan: Can you remove any `console.log` statements?

client/src/components/CarouselDots/CarouselDots.js
```diff
@@ -0,0 +1,91 @@
...
+
+    this.state = {
+      classList: "CarouselDots",
+      activeLocation: props.activeLocation,
+      activeFill: "#6A4C93",
+      inactiveFill: "#FAFAFA"
```
ZakkMan: Could we set this with CSS somehow? Then we could use the CSS variables

client/src/components/CarouselDots/CarouselDots.js
```diff
@@ -0,0 +1,91 @@
...
+    };
+  }
+  componentDidUpdate(props) {
+    console.log("Carousel Updated");
+  }
+
```
ZakkMan: Remove please!

client/src/components/CarouselDots/CarouselDots.js
```diff
@@ -0,0 +1,91 @@
...
+  dotClicked = index => {
```
ZakkMan: Can we give this a name like `handleClick`

client/src/components/CarouselDots/CarouselDots.js
```diff
@@ -0,0 +1,91 @@
...
+    this.setState({ activeLocation: index }, () => {
+      this.props.onStatusChange("Updated");
+    });
+  };
+
+  componentWillUnmount() {}
```
ZakkMan: Remove if you're not using this

client/src/components/CarouselDots/CarouselDots.js
```diff
@@ -0,0 +1,91 @@
...
+            this.dotClicked(i);
+          }}
+          cx={proportions / 2}
+          cy={proportions / 2}
+          r={proportions / 5}
+          fill={fill}
+        />
+      </svg>
+    );
+  };
```
ZakkMan: Can you condense this to a single ternary in the `render` method?

client/src/components/CarouselDots/CarouselDots.js
```diff
@@ -0,0 +1,91 @@
...
+
+  render() {
+    let dots = this.generateDots();
+    return (
+      <div className={this.state.classList} style={this.state.style}>
+        {dots}
+      </div>
+    );
+  }
+}
```
ZakkMan: Can you put everything in `render`?

client/src/components/CarouselDots/CarouselDots.stories.js
```diff
@@ -0,0 +1,28 @@
...
+  backgroundColor: "#FAFAFA"
+};
+
+let location = 1;
+
+storiesOf("CarouselDots", module)
+  .add("Onboarding Step 2", () => (
+    <div style={style1}>
+      <CarouselDots numberOfDots={5} activeLocation={location} onStatusChange={statusChange} />
+    </div>
```
ZakkMan: I think these divs with the backgrounds should maybe be part of the component itself

client/src/components/CarouselDots/CarouselDots.css
```diff
@@ -0,0 +1,28 @@
...
+
+
+.CarouselDot__Border {
+  border: 2px solid var(--purple);
+}
+
+.CarouselDot__Active {
+  background-color: var(--purple)
+}
```
ZakkMan: Modifier/element names should be lower case (i.e. `CarouselDot__border`). Also, `active` is more of a modifier - so maybe `.CarouselDot--active`

client/src/components/CarouselDots/CarouselDots.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  dotClicked = evt => {
+    let index = evt.target.id;
+    this.setState({ activeLocation: index }, () => {
+      this.props.onStatusChange(index);
+    });
+  };
+
+  componentWillUnmount() {}
```
ZakkMan: Can you delete this, since you're not using it?

client/src/components/CarouselDots/CarouselDots.js
```diff
@@ -0,0 +1,56 @@
...
+
+  generateClassList(active) {
+    let classList = "CarouselDot ";
+    classList = this.props.hasOwnProperty("border")
+      ? (classList += " CarouselDot__Border ")
+      : classList;
+    classList = active ? (classList += "CarouselDot__Active ") : classList;
+    return classList;
+  }
```
ZakkMan: Can you find a cleaner way of doing this inside of `render`?

client/src/components/CarouselDots/CarouselDots.js
```diff
@@ -0,0 +1,56 @@
...
+class CarouselDots extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      classList: "CarouselDots",
+      activeLocation: props.activeLocation
+    };
+  }
+
+  dotClicked = evt => {
```
ZakkMan: Let's call this method something like `handleClick`

client/src/components/CarouselDots/CarouselDots.stories.js
```diff
@@ -0,0 +1,34 @@
...
+let location = 1;
+
+storiesOf("CarouselDots", module)
+  .add("Onboarding Step 2", () => (
+    <div style={style1}>
+      <CarouselDots
+        numberOfDots={5}
+        activeLocation={location}
+        onStatusChange={statusChange} />
+    </div>
```
ZakkMan: The background should be part of the component

client/src/components/CarouselDots/CarouselDots.js
```diff
@@ -0,0 +1,58 @@
...
+  };
+
+  generateClassList(active) {
+    let classList = "CarouselDot ";
+    classList = this.props.hasOwnProperty("border")
+      ? (classList += " CarouselDot__border ")
+      : classList;
+    classList = active ? (classList += "CarouselDot--active ") : classList;
+    return classList;
+  }
```
ZakkMan: Are you still using this? If not, then can you remove it?

client/src/components/CarouselDots/CarouselDots.js
```diff
@@ -0,0 +1,58 @@
...
+
+  render() {
+    let numberOfDots = this.props.numberOfDots;
+    let activeLocation = this.state.activeLocation;
+    console.log(activeLocation);
```
ZakkMan: Can you remove any `console.log` statements in your code?

Issue #75, opened 2019-12-06
Component: Button/Input Type Submit Component Issue


clarknoah:
<img width="297" alt="Screen Shot 2019-12-06 at 2 19 35 PM" src="https://user-images.githubusercontent.com/1622159/70350022-e6097680-1833-11ea-84aa-f536a2fc6022.png">

<img width="183" alt="Screen Shot 2019-12-06 at 2 17 59 PM" src="https://user-images.githubusercontent.com/1622159/70350041-ee61b180-1833-11ea-92a7-3fdaef3f4742.png">

<img width="287" alt="Screen Shot 2019-12-06 at 2 19 11 PM" src="https://user-images.githubusercontent.com/1622159/70350032-eb66c100-1833-11ea-82f9-fb7e39de6ece.png">

<img width="287" alt="Screen Shot 2019-12-06 at 2 19 21 PM" src="https://user-images.githubusercontent.com/1622159/70350027-e86bd080-1833-11ea-90cf-44bfbf213d8c.png">

- Location:
    - Onboarding
    - #PackingLight
    - Pretty much everywhere!
- Features
    - Clickable button that should allow the parent to run a function each time it is clicked
    - It could be button or input type submit
- Inputs
    - backgroundColor={
    - TextColor
- Internal Actions
    - Style correctly based on input properties
    - buttonClicked function which should execute the `this.props.clickedAction` function
- Outputs
    - Component should have clickedAction={parentFunctionClicked} which accepts an object from within the com

karenheyn:
move to #168

Issue #16, opened 2019-12-05
Component: Button Issue


tissakhosla:
![image](https://user-images.githubusercontent.com/53238865/70267423-711f3980-176c-11ea-84b9-5843b0b8bf56.png)

![image](https://user-images.githubusercontent.com/53238865/70269902-4daabd80-1771-11ea-8f72-350f10c771fd.png)

![image](https://user-images.githubusercontent.com/53238865/70270065-8e0a3b80-1771-11ea-9f69-44ec31fb0112.png)

Button Component.

Properties:

- ```text```: string
- ```onClick```: callback

MobolanleAdebesin:
Button Component. 
Properties: 
- text: string 
- onClick: callback

llucas314:
moved to #68

PR #112, opened 2019-12-07
Component(s): Buttonsubmit

karenheyn:
I made all the submit buttons as per issue #75
all buttons have an onClick prop so they can call a function to handle submits.

### DIFF
--- a/client/src/components/ButtonSubmit/ButtonSubmit.css
+++ b/client/src/components/ButtonSubmit/ButtonSubmit.css
@@ -1,0 +1,22 @@
+@import url('https://fonts.googleapis.com/css?family=Reem+Kufi&display=swap');
+
+
+.ButtonSubmit-followed,
+.ButtonSubmit-favorited {
+  background: #FB836F;
+  box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
+  border-radius: 4px;
+  width: 200px;
+  height: 50px;
+  color: white;
+  font-family: 'Reem Kufi', sans-serif;
+  font-size: 18px;
+}
+
+.ButtonSubmit-next {
+  width: 302px;
+  height: 65px;
+  background: #FAFAFA;
+  box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
+  border-radius: 3px;
+}
--- a/client/src/components/ButtonSubmit/ButtonSubmit.css
+++ b/client/src/components/ButtonSubmit/ButtonSubmit.css
@@ -8,15 +8,118 @@
   border-radius: 4px;
   width: 200px;
   height: 50px;
-  color: white;
+  color: #FAFAFA;
   font-family: 'Reem Kufi', sans-serif;
   font-size: 18px;
 }
 
-.ButtonSubmit-next {
+.ButtonSubmit-next,
+.ButtonSubmit-finish {
   width: 302px;
   height: 65px;
   background: #FAFAFA;
   box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
   border-radius: 3px;
+  font-family: 'Reem Kufi', sans-serif;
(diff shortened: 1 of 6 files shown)


client/src/components/ButtonSubmit/ButtonSubmit.css
```diff
@@ -0,0 +1,107 @@
+@import url('https://fonts.googleapis.com/css?family=Reem+Kufi&display=swap');
+
+.ButtonSubmit {
+  box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
+  font-family: 'Reem Kufi', sans-serif;
```
ZakkMan: Can you use the CSS variable for the font name?

client/src/components/ButtonSubmit/ButtonSubmit.css
```diff
@@ -0,0 +1,107 @@
...
+}
+
+.ButtonSubmit-followed,
+.ButtonSubmit-favorited {
+  background: #FB836F;
```
ZakkMan: Can you use the CSS variables for the colors too?

client/src/components/ButtonSubmit/ButtonSubmit.css
```diff
@@ -0,0 +1,107 @@
...
+  border-radius: 4px;
+  width: 200px;
+  height: 50px;
+  color: #FAFAFA;
+  font-size: 18px;
+}
```
ZakkMan: Rather than giving these an explicit height and width, we want these to have padding ratio that is consistent across all the button types.

We may also want to give these a `max-width` property incase the text is really long for some reason

client/src/components/ButtonSubmit/ButtonSubmit.css
```diff
@@ -0,0 +1,107 @@
...
+  background: #FAFAFA;
+  border-radius: 3px;
+  font-size: 36px;
+}
+
+.ButtonSubmit-register {
+  width: 302px;
+  height: 65px;
+  background: #6A4C93;
+  border-radius: 5px;
```
ZakkMan: Can you pick a single `border-radius` value and add that to the base button class?

client/src/components/ButtonSubmit/ButtonSubmit.css
```diff
@@ -0,0 +1,107 @@
...
+  color: #FAFAFA;
+  font-size: 36px;
+}
+
+.ButtonSubmit-login {
+  width: 302px;
+  height: 65px;
+  background: #FAFAFA;
+  border-radius: 5px;
+  font-size: 36px;
```
ZakkMan: It looks like there are two `font-size` values - `18px` and `36px`. Could we have these be part of small and large classes that also affect the padding?

client/src/components/ButtonSubmit/ButtonSubmit.js
```diff
@@ -0,0 +1,29 @@
...
+    "login",
+    "joinTribe",
+    "register",
+    "connect",
+    "message",
+    "join",
+    "book",
+    "follow",
+    "post"
+  ];
```
ZakkMan: There aren't really different types of buttons - just some different styles. Like the "followed" and "favorited" buttons are the same, they just have different text.

client/src/components/ButtonSubmit/ButtonSubmit.js
```diff
@@ -0,0 +1,29 @@
...
+  if (types.includes(props.type)) {
+    classList += ` ${classList}-${props.type}`;
+  }
+  return <button className={classList}>{props.label && <span>{props.label} </span>}</button>;
```
ZakkMan: This is really clean and looks really good

shelbyvjacobs:
This is a duplicate of PR #115 / Issue #16

Issue #18, opened 2019-12-05
Component: Footer Issue


tissakhosla:
![image](https://user-images.githubusercontent.com/53238865/70267666-e559dd00-176c-11ea-9af0-e91716778e60.png)

These are links for the footer. Some are internal and some are external. 

If an external link, it will open in a new page/tab. 

```
position: absolute;
left: 0%;
right: 0%;
top: 100%;
bottom: -100%;
height: 63px;

background: #FFCB08;
transform: matrix(1, 0, 0, -1, 0, 0);
```

wchrapcynski:
These are links for the footer. Some are internal and some are external. 

Properties:
* `text`: string
* `link`: string
* `onStatusChange`: callback

If an external link, it will open in a new page/tab. Otherwise it will have a callback for loading that component.

ZakkMan:
@tissakhosla Do we need those properties? We'll probably just render the `<Footer />` component without any properties.

PR #114, opened 2019-12-07
Footer component PR

wchrapcynski:
Footer component is complete for issue #18. Some of the CSS will need to be adjusted once the variables are merged in for use. Also, the links in the footer will just been hard links. Those will need to be updated as information becomes available.


ZakkMan:
@wchrapcynski Make sure you reference the issue in your PRs

client/src/components/Footer/Footer.css
```diff
@@ -0,0 +1,47 @@
+@import url('https://fonts.googleapis.com/css?family=Reem+Kufi&display=swap'); /* Will eventually delete */
```
ZakkMan: This should be set globally, in `App.css`

client/src/components/Footer/Footer.css
```diff
@@ -0,0 +1,47 @@
...
+
+.Footer {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    min-width: 100vw;
+    height: 50px;
+    background-color: #FFCB06; /* Change colors once css variables are merged */
+    font-family: 'Reem Kufi', sans-serif; /* Will eventually delete */
```
ZakkMan: The variables are merged, so you should be good to use those!

client/src/components/Footer/Footer.css
```diff
@@ -0,0 +1,47 @@
...
+    .Footer {
+        height: 75px;
+    }
+
+    .footer-right {
+        flex-direction: column;
+        justify-content: right;
+        line-height: 20px;
+    }
+}
```
ZakkMan: This is a great start to responsive! On mobile can we have all three columns stack?

wchrapcynski: @ZakkMan Made edits and pushed them up. The footer collapses to a single column below 400px. All should be ready.

client/src/components/Footer/Footer.css
wchrapcynski: Updated the component CSS to reflect this.

PR #115, opened 2019-12-07
Component: Button PR

llucas314:
The button component has the properties of text, size, color, and handleClick. All button variations in the mockup can be produced from this component.

This pull request is for issue #16


client/src/components/Button/Button.css
```diff
@@ -0,0 +1,61 @@
...
+.btn-pink:hover {
+  background-color: #d42e44;
+}
+.btn-gold {
+  color: #fafafa;
+  background-color: #ffcb08;
+}
+.btn-gold:hover {
+  background-color: #f5c000;
+}
```
ZakkMan: We added CSS variables for the colors. Can you switch this to use those?

client/src/components/Button/Button.css
```diff
@@ -0,0 +1,61 @@
+@import url("https://fonts.googleapis.com/css?family=Reem+Kufi&display=swap");
+
+.Button {
+  width: 200px;
+  height: 50px;
```
ZakkMan: Rather than setting explicit height and width values, can you assign padding ratios? That way, these will look consistent when the text is short and when it's long

client/src/components/Button/Button.css
```diff
@@ -0,0 +1,61 @@
...
+  background: #fafafa;
+  box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
+  border-radius: 5px;
+  font-family: "Reem Kufi", sans-serif;
```
ZakkMan: We have CSS variables now for font and color properties! Can you use those?

client/src/components/Button/Button.css
```diff
@@ -0,0 +1,61 @@
...
+  color: #252525;
+  -webkit-transition-duration: 0.3s;
+  transition-duration: 0.3s;
+  -webkit-transition-property: background-color;
+  transition-property: background-color;
+}
+.Button:hover {
+  background-color: #fcf0ee;
+}
+.btn-large {
```
ZakkMan: If the base class is going to be `button`, then we should be consistent and use that throughout. Can you change these to `button-large`, `button-small`, etc?

llucas314:
@ZakkMan  I updated the button to use the CSS variables, used the padding ratio for the different sizes and renamed the classes.

llucas314:
Moved to #113

llucas314:
split pill and button from #113

client/src/components/Button/Button.js
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import "./Button.css";
+
+const Button = props => {
+  Button.defaultProps = {
+    color: "color",
+    size: "size"
+  };
```
ZakkMan: So cool you're using `defaultProps`! Can you move this out of the function definition though?

llucas314:
@ZakkMan Made requested changes.

Issue #23, opened 2019-12-05
Component: Input Type Text


tissakhosla:

tissakhosla:
![image](https://user-images.githubusercontent.com/55366813/70269431-5ea6ff00-1770-11ea-8ee4-8a71266c6480.png)

TylerEikenberg:
Component complete.

PR #116, opened 2019-12-07
Component: Input Text Field PR

TylerEikenberg:
This component has multiple variations that can all be created by passing in a different value in the 'type' props.  Issue #23. 

I know in the future when we implement this on the site that there will need to be some sort of function to retrieve the data typed into the forms. I wasn't sure how to put that into the component on storybook, but I'd be glad to take care of it whenever we get to actually implementing these forms.


client/src/components/InputTextField/InputTextField.js
```diff
@@ -0,0 +1,143 @@
...
+
+    // Default CSS class to apply to the Component
+    this.state = {
+      classList: "input-text-field-container",
+      classListSmall: "input-text-field-container-small",
+      classListSearch: "input-text-field-container search",
+      email: "",
+      name: ""
+    };
+  }
```
ZakkMan: I don't think this component will need any state - it should take a `onChange` method as a `prop`, but shouldn't manage it's own state. The goal here (of all atomic components) is to just render the piece of the UI. The component rendering it will be responsible for state. Does that make sense?

client/src/components/InputTextField/InputTextField.js
```diff
@@ -0,0 +1,143 @@
...
+              fill="#bababa"
+              d="M9.5,3A6.5,6.5 0 0,1 16,9.5C16,11.11 15.41,12.59 14.44,13.73L14.71,14H15.5L20.5,19L19,20.5L14,15.5V14.71L13.73,14.44C12.59,15.41 11.11,16 9.5,16A6.5,6.5 0 0,1 3,9.5A6.5,6.5 0 0,1 9.5,3M9.5,5C7,5 5,7 5,9.5C5,12 7,14 9.5,14C12,14 14,12 14,9.5C14,7 12,5 9.5,5Z"
+            />
+          </svg>
+          <input type="text" name="name" placeholder="Search Trips" />
+        </div>
+      );
+    }
+  }
+}
```
ZakkMan: Can you simplify this? I don't think we need any of these conditionals. `type`, `name`, and `placeholder`, should all be passed down as props. As should `onChange`. You'll also need a prop for the icon (see #120) and the size

client/src/components/InputTextField/InputTextField.js
TylerEikenberg: Understood, working on changes now. Thanks!

client/src/components/InputTextField/InputTextField.js
TylerEikenberg: Makes sense, refactoring everything now

TylerEikenberg:
@ZakkMan I totally refactored the component. Its now a functional component, and takes in a ```placeholder``` prop, and optional ```size``` and ```type``` props to determine size and if its a search input. Removed a huge chunk of code, looks much cleaner now.

client/src/components/InputTextField/InputTextField.css
```diff
@@ -0,0 +1,41 @@
+.input-text-field-container input {
+  display: flex;
+  color: black;
+  width: 455px;
+  height: 55px;
+  background-color: #fff;
```
ZakkMan: Is there a variable for the color that you can use?

client/src/components/InputTextField/InputTextField.css
```diff
@@ -0,0 +1,41 @@
+.input-text-field-container input {
+  display: flex;
+  color: black;
+  width: 455px;
+  height: 55px;
```
ZakkMan: It'd be better to setting padding here to get a consistent sizing rather than to set an explicit height and width

client/src/components/InputTextField/InputTextField.css
```diff
@@ -0,0 +1,41 @@
...
+  border-radius: 3px;
+  padding: 0px 30px;
+  font-size: 20px;
+  font-family: var(--default-font);
+}
+
+.input-text-field-container-small input {
+  color: black;
+  width: 175px;
+  height: 42px;
```
ZakkMan: Dito here re padding

client/src/components/InputTextField/InputTextField.css
```diff
@@ -0,0 +1,41 @@
...
+  background-color: #fff;
+  /* border: 2px solid black; */
```
ZakkMan: Delete commented out code!

client/src/components/InputTextField/InputTextField.js
```diff
@@ -0,0 +1,40 @@
...
+import "./InputTextField.css";
+import "../../App.css";
+/**
+ * Created by Tyler
+ *
+ * Props will define the type of input field
+ * Input fields include email, password, name, confirm password, city of residence, what is your age?, what is your dream destination, First name,   last name, search my directory, search trips.
+ *
+ * I think we'll need to implement a function on these to retrieve the data typed in when pressing a submit button. I wasn't sure how to do this with just the standalone components in storybook.
+ */
```
ZakkMan: This documentation is great! We're going to figure out a way to provide this kind of component documentation so you don't have to do this

client/src/components/InputTextField/InputTextField.js
```diff
@@ -0,0 +1,40 @@
...
+const InputTextField = props => {
+  let { size, type, name, placeholder, onChange } = props;
+  let classList = "input-text-field-container";
+  if (size === "small") {
+    classList = "input-text-field-container-small";
+  } else if (type === "search") {
+    classList = "input-text-field-container search";
+  } else if (size === "small" && type === "search") {
+    classList = "input-text-field-container-small search";
+  }
```
ZakkMan: I'd set a default prop value for `type` (maybe `"text"`) - that way you know you'll always have a value. Then you can modify this to just be a single line of string interpolation:

```js
let classList = `input-text-field-container intput-text-field-container-${type}`
```

If `type` is `email`, then you'll get: `"input-text-field-container intput-text-field-container-email"`

Another side note is these classnames are super long. You could go with `input-field`, `input-field-container`, `input-field-text`, `input-field-email`, `input-field-search`, etc

ZakkMan: Oh and better yet - don't do this with a variable, just do it on line 26 when you set the `className` for the `div`

client/src/components/InputTextField/InputTextField.css
wchrapcynski: There should be. Pure white or black isn't actually used. So, the variables for black or white will work in their place.

TylerEikenberg:
@ZakkMan

client/src/components/InputTextField/InputTextField.js
```diff
@@ -0,0 +1,26 @@
...
+    <div className={classList}>
+      {type === "search" ? (
+        <svg viewBox="-4 -2 24 24">
+          <path
+            fill="#bababa"
+            d="M9.5,3A6.5,6.5 0 0,1 16,9.5C16,11.11 15.41,12.59 14.44,13.73L14.71,14H15.5L20.5,19L19,20.5L14,15.5V14.71L13.73,14.44C12.59,15.41 11.11,16 9.5,16A6.5,6.5 0 0,1 3,9.5A6.5,6.5 0 0,1 9.5,3M9.5,5C7,5 5,7 5,9.5C5,12 7,14 9.5,14C12,14 14,12 14,9.5C14,7 12,5 9.5,5Z"
+          />
+        </svg>
+      ) : null}
+      <input type="text" name={name} placeholder={placeholder} onChange={onChange} />
```
ZakkMan: The `type` attribute is hard coded to be `text`, but we'll want it to be `email` and `password` sometimes. Some of your stories are passing in a `type` prop for this already

TylerEikenberg: @ZakkMan added prop for input type. It's getting a little prop heavy but hopefully this isn't going overboard?

client/src/components/InputTextField/InputTextField.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import "./InputTextField.css";
+
+const InputTextField = props => {
+  let { type, name, placeholder, onChange } = props;
+  let classList = `input-field-container-${type}`;
+  InputTextField.defaultProps = {
+    type: "default"
+  };
```
ZakkMan: Can you move this to outside the component definition?

Read this for more details: https://blog.logrocket.com/a-complete-guide-to-default-props-in-react-984ea8e6972d/

client/src/components/InputTextField/InputTextField.js
```diff
@@ -0,0 +1,26 @@
...
+          <path
+            fill="#bababa"
+            d="M9.5,3A6.5,6.5 0 0,1 16,9.5C16,11.11 15.41,12.59 14.44,13.73L14.71,14H15.5L20.5,19L19,20.5L14,15.5V14.71L13.73,14.44C12.59,15.41 11.11,16 9.5,16A6.5,6.5 0 0,1 3,9.5A6.5,6.5 0 0,1 9.5,3M9.5,5C7,5 5,7 5,9.5C5,12 7,14 9.5,14C12,14 14,12 14,9.5C14,7 12,5 9.5,5Z"
+          />
+        </svg>
+      ) : null}
+      <input type={props.input} name={name} placeholder={placeholder} onChange={onChange} />
+    </div>
+  );
+};
```
ZakkMan: Can we use the `type` property for the `type` attribute on the `input` element and rename `type` (i.e. the type of input component) to something else?

client/src/components/InputTextField/InputTextField.js
TylerEikenberg: Added default value to destructor set variation to "default" by default.

client/src/components/InputTextField/InputTextField.js
TylerEikenberg: Switched everything around per your comment

TylerEikenberg:
@ZakkMan

Issue #52, opened 2019-12-06
Component: Welcome Header


TylerEikenberg:
<img width="1434" alt="Screen Shot 2019-12-06 at 12 04 08 PM" src="https://user-images.githubusercontent.com/55251251/70348228-44345a80-1830-11ea-81a0-b985e9b48461.png">

**Props:**
- Users first name

ZakkMan:
This could probably be merged with #71 and #97

TylerEikenberg:
Merging this component with #71 and #97

Issue #97, opened 2019-12-06
Component: Trip Header


stuvun:
**Location: Hawaii 2020**
![banner](https://user-images.githubusercontent.com/55634628/70352722-c7a67980-1839-11ea-8a3d-23ac92442e2d.png)
![travelBanner](https://user-images.githubusercontent.com/55634628/70352728-ca08d380-1839-11ea-862a-16a42b4fbd77.png)
**Props:**
* `location`

**Comments:**
Initially, I had these as two separate components, but I thought it would be more simple to combine these two as one component because they'd both be using the same prop. (The text would be the `location` prop and the image for the bottom half would be dictated by the `location` prop as well)

MobolanleAdebesin:
Closing this issue as it is being addressed in issue #149

Issue #71, opened 2019-12-06
Component: Molecular Pink Header Issue


torykling:
![image](https://user-images.githubusercontent.com/55210397/70350359-9ecfb580-1834-11ea-9fa3-2e6f297347f4.png)

Props:
-Title

Atom component: 
-Input Type Text

Locations: 
-Members Profile
-Forum Discover
-Directory People
-Directory Groups

```
/* cotripper_banner2-01 */


position: absolute;
width: 1440px;
height: 231px;
left: -2px;
top: 128px;

background: url(cotripper_banner2-01.png);
box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
```

bmillsmc:
If you are on this issue please move to issue #164

PR #117, opened 2019-12-07
Component: Welcome header PR (repeat component)

TylerEikenberg:
Pretty straightforward for this one. Going by the mockup, props will be the users first name so the storybook example shows "Welcome, Tyler!" and "Welcome, Example!". 
The font is default for now. #52

### DIFF
--- a/client/src/components/WelcomeHeader/WelcomeHeader.css
+++ b/client/src/components/WelcomeHeader/WelcomeHeader.css
@@ -14,6 +14,10 @@
   font-family: var(--default-font);
 }
 
+.community {
+  background-image: var(--splash-background);
+}
+
 @media only screen and (min-width: 768px) {
   .welcome-header {
     font-size: 90px;
--- a/client/src/components/WelcomeHeader/WelcomeHeader.css
+++ b/client/src/components/WelcomeHeader/WelcomeHeader.css
@@ -1,13 +1,14 @@
 .welcome-header {
   width: 1440px;
-  height: 380px;
-  background-image: url("../../../../assets/images/home_banner.png");
+  min-height: 380px;
+  height: 30vh;
+  background-image: var(--home-banner);
   -webkit-box-shadow: 0px 5px 5px 0px rgba(0, 0, 0, 0.4);
   -moz-box-shadow: 0px 5px 5px 0px rgba(0, 0, 0, 0.4);
   box-shadow: 0px 5px 5px 0px rgba(0, 0, 0, 0.4);
   text-align: center;
   line-height: 400%;
-  color: white;
+  color: var(--white);
   font-size: 90px;
-  /* font-family: ???; */
+  font-family: var(--default-font);
 }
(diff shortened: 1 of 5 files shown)


client/src/components/WelcomeHeader/WelcomeHeader.js
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+import "./WelcomeHeader.css";
+
+// Created by Tyler. Not sure what font we're going with yet so feel free to adjust that and play around with the font size.
+const WelcomeHeader = props => {
+  // Default Class to apply to Component
+  let classList = `welcome-header`;
+
+  return <div className={classList}>Welcome, {props.name}!</div>;
```
ZakkMan: If you have this render children (i.e. `{children}`), then this could handle #52, #71, and #97 

That would modify the use slightly. You would render:

```jsx
<Header>
Welcome, {this.props.user.name}
</Header>
```

But then for #97, we could do something like this:

```jsx
<Header>
  <h3>Trip: Hawaii 2020</h3>
</Header>
```

and for #71, we could do this:

```jsx
<Header>
  <h3>User Profile</h3>
  <SearchInput placeholder={"Search My Directory"} onSubmit={this.handleSumbit} />
</Header>
```

client/src/components/WelcomeHeader/WelcomeHeader.css
```diff
@@ -0,0 +1,13 @@
...
+  height: 380px;
+  background-image: url("../../../../assets/images/home_banner.png");
+  -webkit-box-shadow: 0px 5px 5px 0px rgba(0, 0, 0, 0.4);
+  -moz-box-shadow: 0px 5px 5px 0px rgba(0, 0, 0, 0.4);
+  box-shadow: 0px 5px 5px 0px rgba(0, 0, 0, 0.4);
+  text-align: center;
+  line-height: 400%;
+  color: white;
+  font-size: 90px;
+  /* font-family: ???; */
```
ZakkMan: You can use the CSS Variable for the font and color settings now!

client/src/components/WelcomeHeader/WelcomeHeader.css
```diff
@@ -0,0 +1,13 @@
+.welcome-header {
+  width: 1440px;
+  height: 380px;
```
ZakkMan: This would be a great place to use `vh` and `vw`. We always want the header to be `100vh`. We'll want it to have a `min-height`, which could be `380px`. Otherwise, we'll want it to be some percentage of the view height (maybe 25 or 30 - what ever you think looks best!).

TylerEikenberg: Could you clarify how the width with vw would be handled? Setting the width to 100vw look s like this, with the background image repeating I just thought it looked weird.
Screen Shot 2019-12-08 at 1 42 47 PM

TylerEikenberg:
Stopping work on this component because they're the same as #71 and #97 
@cenwachukwu @MobolanleAdebesin

vmipfdr:
It looks like #97 is the same as this one, #117 @ZakkMan @sarahtsige

Issue #91, opened 2019-12-06
Component: Activity List


stuvun:
**Location: Hawaii 2020**
![momActivities](https://user-images.githubusercontent.com/55634628/70351730-8f9e3700-1837-11ea-9af6-163c78f53021.png)
![momKidActivities](https://user-images.githubusercontent.com/55634628/70351731-90cf6400-1837-11ea-955a-e87a4a4e4fc4.png)
**Props:**
* `location`

stuvun:
Here's a link to the pull request:
[Pull Request](https://github.com/CotripperPlatform/CoTrip/pull/119)

PR #119, opened 2019-12-07
Component: ActivityList

stuvun:
[Link to issue](https://github.com/CotripperPlatform/CoTrip/issues/91)
Please let me know if I formatted this correctly and did the media queries right. I wasn't sure if there was a standard set for how to do this, so I just went with this for now.


ZakkMan:
@stuvun Make sure you assign a reviewer and assign yourself the PR

client/src/components/ActivityList/ActivityList.css
```diff
@@ -0,0 +1,105 @@
+@import url('https://fonts.googleapis.com/css?family=Reem+Kufi&display=swap');
+
+.ActivityList {
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+    padding: 0px 40px;
+}
+
+.activitiesContainer {
```
ZakkMan: We want to have consistent text formatting for our CSS classes. Could you make them all kebab case?

client/src/components/ActivityList/ActivityList.css
```diff
@@ -0,0 +1,105 @@
...
+    padding: 20px 0px;
+}
+
+.activityHeader {
+    font-family: 'Reem Kufi', sans-serif;
+    font-style: normal;
+    font-weight: normal;
```
ZakkMan: Can you use the CSS variables for this?

client/src/components/ActivityList/ActivityList.css
```diff
@@ -0,0 +1,105 @@
...
+    font-size: 36px;
+    line-height: 54px;
```
ZakkMan: Maybe use a heading tag here? That will cut down on the amount of styling you need here.

client/src/components/ActivityList/ActivityList.css
```diff
@@ -0,0 +1,105 @@
...
+    text-align: center;
+}
+
+.mom {
+    background: #6A4C93;
+}
+
+.kids {
+    background: #D1495B;
+}
```
ZakkMan: Rather than give these explicit classes and tie the color to the content, accept a prop in the `Activity` Component that determines the color

client/src/components/ActivityList/ActivityList.css
```diff
@@ -0,0 +1,105 @@
...
+
+@media all and (max-width: 1024px) and (min-width: 721px) {
+    .activityList {
+        width: 75vw;
```
ZakkMan: I like how you're thinking here, but it'd be better to use a percentage here

client/src/components/ActivityList/ActivityList.css
```diff
@@ -0,0 +1,105 @@
...
+    }
+    .activityHeader {
+        font-size: 27px;
+        line-height: 40.5px;
+    }
+    .activity {
+        width: 135px;
+        height: 101.25px;
+        margin: 30px 30px 15px 0px;
+        padding: 7.5px;
```
ZakkMan: There's no such thing as a half pixel!

client/src/components/ActivityList/ActivityList.js
```diff
@@ -0,0 +1,73 @@
+import React, {Component} from 'react';
+import "./ActivityList.css";
+
+// Class Based React Component
+class ActivityList extends Component{
```
ZakkMan: Both `Activity` and `ActivityList` should be function components, there wont be any state that either needs to manage.

client/src/components/ActivityList/ActivityList.js
```diff
@@ -0,0 +1,73 @@
...
+  renderMomActivities = () => {
+    const momActivityList = this.momActivityList
+    let momActivities = [];
+    momActivityList.forEach(activity => {
+      momActivities.push(
+        <div className='activity mom'>
+          <span className='activities'>{activity}</span>
+        </div>
+      )
+    })
```
ZakkMan: If you're going to loop and create a new array, then you should use `.map()`

client/src/components/ActivityList/ActivityList.js
```diff
@@ -0,0 +1,73 @@
...
+  renderKidsActivities = () => {
+    const kidsActivityList = this.kidsActivityList
+    let kidsActivities = [];
+    kidsActivityList.forEach(activity => {
+      kidsActivities.push(
+        <div className='activity kids'>
+          <span className='activities'>{activity}</span>
+        </div>
+      )
+    })
```
ZakkMan: Ditto here - use `.map()`

client/src/components/ActivityList/ActivityList.js
```diff
@@ -0,0 +1,73 @@
...
+  // Runs after a component has been updated
+  componentDidUpdate(){}
+
+
+  // Runs right before a component is removed from the DOM
+  componentWillUnmount(){}
+  
+  render(){
+    const renderMomActivities = this.renderMomActivities()
+    const renderKidsActivities = this.renderKidsActivities()
```
ZakkMan: I like how you're thinking here, but it's better to do this inline in the JSX rather than in a separate function

stuvun:
@ZakkMan Sorry, just read this. Thanks for all the suggestions! I'll try to fix them tomorrow.

client/src/components/ActivityList/ActivityList.js
```diff
@@ -0,0 +1,72 @@
...
+    this.momActivityList = ["Spa", "Wine Nights", "Brunch with Other Moms"];
+    this.kidsActivityList = [
+      "Snorkeling",
+      "Hiking",
+      "Island Hopping",
+      "Hawaiian Luaus",
+      "Food Markets",
+      "Beaches"
+    ];
+  }
```
ZakkMan: This data should be passed in as props

client/src/components/ActivityList/ActivityList.js
```diff
@@ -0,0 +1,72 @@
+import React, { Component } from "react";
+import "./ActivityList.css";
+
+// Class Based React Component
+class ActivityList extends Component {
+  constructor(props) {
+    super(props);
+    console.log(props);
```
ZakkMan: Remove any `console.log` statements

client/src/components/ActivityList/ActivityList.js
```diff
@@ -0,0 +1,72 @@
...
+
+    // Default CSS class to apply to the Component
+    this.state = {
+      classList: "ActivityList__activity-list"
+    };
```
ZakkMan: You don't need to track the classList in state - just assign it to the `className` attribute

client/src/components/ActivityList/ActivityList.js
```diff
@@ -0,0 +1,72 @@
...
+  renderKidsActivities = () => {
+    const kidsActivities = this.kidsActivityList.map(activity => {
+      return (
+        <div className="ActivityList__activity ActivityList__kids">
+          <span className="ActivityList__activities">{activity}</span>
+        </div>
+      );
+    });
+    return <div className="ActivityList__activity-container">{kidsActivities}</div>;
+  };
```
ZakkMan: Can you move this into the `render` method?

client/src/components/ActivityList/ActivityList.js
```diff
@@ -0,0 +1,72 @@
...
+
+  // Runs after Component is loaded in the broswer
+  componentDidMount() {}
+
+  // Runs after a component has been updated
+  componentDidUpdate() {}
+
+  // Runs right before a component is removed from the DOM
+  componentWillUnmount() {}
```
ZakkMan: Can you delete these since you're not using them?

client/src/components/ActivityList/ActivityList.js
```diff
@@ -0,0 +1,72 @@
...
+        <div className="ActivityList__activities-container">
+          <h1 className="ActivityList__activity-header">Activities for Mom</h1>
+          {this.renderMomActivities()}
+        </div>
+        <div className="ActivityList__activities-container">
+          <h1 className="ActivityList__activity-header">Activities for Mom & Kids</h1>
+          {this.renderKidsActivities()}
+        </div>
+      </div>
+    );
```
ZakkMan: This should be two instances of the `ActivityList` component. One with the text "Activities for Moms" and another with text "Activities for Moms & Kids"

client/src/components/ActivityList/ActivityList.js
```diff
@@ -0,0 +1,72 @@
...
+    });
+    return <div className="ActivityList__activity-container">{momActivities}</div>;
+  };
+
+  renderKidsActivities = () => {
+    const kidsActivities = this.kidsActivityList.map(activity => {
+      return (
+        <div className="ActivityList__activity ActivityList__kids">
+          <span className="ActivityList__activities">{activity}</span>
+        </div>
```
ZakkMan: This should use the `Card` component

stuvun: A Card component for the ActivityList component wasn't created because the component became orphaned. Since the ActivityList cards were unique to that specific page and there weren't any pages of activities to link these activity cards to, we agreed that this would just be it's own styled text list of activities.

client/src/components/ActivityList/ActivityList.stories.js
```diff
@@ -0,0 +1,25 @@
...
+  "Hiking",
+  "Island Hopping",
+  "Hawaiian Luaus",
+  "Food Markets",
+  "Beaches"
+];
+
+storiesOf("ActivityList", module)
+  .add("Activities for Moms", () => (
+    <ActivityList type="mom" label="Activities for Moms" momActivityList={momActivityList} />
```
ZakkMan: Can you get rid of the `type` property? There isn't really a `type` of activity list - just an activity list with different text for the header and a different color

client/src/components/ActivityList/ActivityList.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  render() {
+    const renderActivities = () => {
+      if (this.props.type === "mom") {
+        const momActivities = this.props.momActivityList.map((activity, index) => {
+          return (
+            <div key={index} className="ActivityList__activity ActivityList__mom">
+              <span className="ActivityList__activities">{activity}</span>
+            </div>
```
ZakkMan: This should use the `Card` component

client/src/components/ActivityList/ActivityList.js
```diff
@@ -0,0 +1,46 @@
...
+        const kidsActivities = this.props.kidsActivityList.map((activity, index) => {
+          return (
+            <div key={index} className="ActivityList__activity ActivityList__kids">
+              <span className="ActivityList__activities">{activity}</span>
+            </div>
+          );
+        });
+        return <div className="ActivityList__activity-container">{kidsActivities}</div>;
+      }
+    };
```
ZakkMan: Can you refactor this code down and remove it from this function?

client/src/components/ActivityList/ActivityList.js
```diff
@@ -0,0 +1,46 @@
+import React, { Component } from "react";
+import "./ActivityList.css";
+
+class ActivityList extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      classList: "ActivityList"
+    };
```
ZakkMan: Can you remove the `classList` from state and turn this into a function component

client/src/components/ActivityList/ActivityList.stories.js
ZakkMan: That said, can you add a `color` property that controls the color?

client/src/components/ActivityList/ActivityList.stories.js
```diff
@@ -0,0 +1,25 @@
...
+  ))
+  .add("Activities for Moms & Kids", () => (
+    <ActivityList
+      type="kids"
+      label="Activities for Moms & Kids"
+      kidsActivityList={kidsActivityList}
+    />
+  ));
```
ZakkMan: There should just be one `activities` prop that accepts an array of strings - not a `momActivityList` and `kidsActivityList`

PR #121, opened 2019-12-07
Modal Components

cenwachukwu:
1. I added a this.prop.children for adding the name of the group user wants to join or the person they want to follow.


ZakkMan:
@cenwachukwu make sure you assign yourself the PR

client/src/components/Modal/Modal.css
```diff
@@ -0,0 +1,73 @@
+@import url('https://fonts.googleapis.com/css?family=Reem+Kufi&display=swap');
```
ZakkMan: This should be set globally

client/src/components/Modal/Modal.css
```diff
@@ -0,0 +1,73 @@
...
+    width: 582px;
+    height: 309.39px;
+    background: #F2BC0F;
+    box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
+    border-radius: 3px;
+    display: flex;
+    flex-direction: column;
+}
+
+.close-modal-btn {
```
ZakkMan: Can you use the full name (i.e. `button` rather than `btn`)

client/src/components/Modal/Modal.css
```diff
@@ -0,0 +1,73 @@
+@import url('https://fonts.googleapis.com/css?family=Reem+Kufi&display=swap');
+
+.Modal {
+    margin: 0;
+    padding: 0;
+    box-sizing: border-box;
+    width: 582px;
+    height: 309.39px;
```
ZakkMan: Rather than set an explicit height and width, can you use percentages and set a the padding?

client/src/components/Modal/Modal.css
```diff
@@ -0,0 +1,73 @@
+@import url('https://fonts.googleapis.com/css?family=Reem+Kufi&display=swap');
+
+.Modal {
+    margin: 0;
+    padding: 0;
+    box-sizing: border-box;
```
ZakkMan: This should be set globally

client/src/components/Modal/Modal.css
```diff
@@ -0,0 +1,73 @@
...
+    width: 582px;
+    height: 309.39px;
+    background: #F2BC0F;
```
ZakkMan: Can you use teh CSS variables for colors and font properties?

client/src/components/Modal/Modal.css
```diff
@@ -0,0 +1,73 @@
...
+    float: right;
+    font-size: 65px;
+    margin-top: -20px;
+}
+
+.modal-body {
+    margin: 0 auto;
+}
+
+.htag{
```
ZakkMan: What does this style? Can you give this a semantic class name?

client/src/components/Modal/Modal.css
```diff
@@ -0,0 +1,73 @@
...
+.modal-footer{
+    margin-top: 75px;
+    height: 67.51px;
+    display: flex;
+    flex-direction: row;
+}
+
+.confirm{
+    border: 1px solid rgba(250, 250, 250, 0.25);
+    width: 285.09px;
```
ZakkMan: There are no fractions on pixels - only absolute numbers

client/src/components/Modal/Modal.js
```diff
@@ -0,0 +1,64 @@
...
+            <button className="cancel">Cancel</button>
+          </div>
+          <div className="confirm">
+            <button className="confirm">Confirm</button>
+          </div>
+        </div>
+      </div>
+    );
+  }
+}
```
ZakkMan: You can shorten this up a lot - this is a lot of duplication. The text of the button can come in as a prop and then. Remember also that you want to accept props for what happens when someone clicks on cancel, the X, or confirm!

client/src/components/Modal/Modal.stories.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import Modal from './Modal';
+
+storiesOf('Modal', module)
+  .add("Follow Modal", () => 
+    <Modal/>
```
ZakkMan: We want to render something like this:

```jsx
<Modal message={"Are you sure you want to Join?"} confirmButtonText={"Join"} onConfirm={this.handleConfirm} />
```

client/src/components/Modal/Modal.css
ZakkMan: @cenwachukwu did you fix this one?

client/src/components/Modal/Modal.css
```diff
@@ -0,0 +1,83 @@
+@import url('https://fonts.googleapis.com/css?family=Roboto&display=swap');
```
ZakkMan: The font is provided globally, so you don't need to import it - just use the CSS variables

client/src/components/Modal/Modal.css
```diff
@@ -0,0 +1,83 @@
...
+
+*{
+    margin: 0;
+    padding: 0;
+    box-sizing: border-box;
+    font-family: var(--default-font);
+}
```
ZakkMan: Can you delete this? It'll be in `App.css`

client/src/components/Modal/Modal.css
```diff
@@ -0,0 +1,83 @@
...
+    width: 582px;
+    height: 309px;
+    background: var(--yellow);
+    box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
+    border-radius: 3px;
+    display: flex;
+    flex-direction: column;
+}
+
+.close-modal-button {
```
ZakkMan: Can you update your class names to be in line with the BEM naming convention?

client/src/components/Modal/Modal.css
```diff
@@ -0,0 +1,83 @@
...
+    color: black;
+    cursor: pointer;
+    float: right;
+    font-family: 'Roboto', sans-serif;
```
ZakkMan: Can you use the variables here for all font properties and color properties?

client/src/components/Modal/Modal.css
```diff
@@ -0,0 +1,83 @@
...
+    box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
+    border-radius: 3px;
+    display: flex;
+    flex-direction: column;
+}
+
+.close-modal-button {
+    color: black;
+    cursor: pointer;
+    float: right;
```
ZakkMan: Can you use flexbox instead of floats?

client/src/components/Modal/Modal.css
```diff
@@ -0,0 +1,83 @@
...
+    background: var(--yellow);
+    box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
+    border-radius: 3px;
+    display: flex;
+    flex-direction: column;
+}
+
+.close-modal-button {
+    color: black;
+    cursor: pointer;
```
ZakkMan: Can you change this to an `a` tag so you don't have to set the cursor manually

client/src/components/Modal/Modal.css
```diff
@@ -0,0 +1,83 @@
...
+    height: 60%;
+}
+
+.modal-message{
+    width: 90%;
+    height: 100%;
+    margin: 0 auto;
+}
+
+h3{
```
ZakkMan: Please don't set styles on tags inside a component - use BEM

client/src/components/Modal/Modal.css
```diff
@@ -0,0 +1,83 @@
...
+    border: 1px solid rgba(37, 37, 37, 0.25);
+    width: 100%;
+    height: 100%;
+    font-size: 24px;
+    line-height: 36px;
+    text-align: center;
+    color: #000000;
+    text-transform: uppercase;
+    background: rgba(37, 37, 37, 0.25);
+}
```
ZakkMan: Are you able to use the `Button` component?

client/src/components/Modal/Modal.js
```diff
@@ -0,0 +1,43 @@
+import React, { Component } from 'react';
+import "./Modal.css";
+import "../../App.css";
```
ZakkMan: Can you delete this?

client/src/components/Modal/Modal.js
```diff
@@ -0,0 +1,43 @@
...
+
+
+// Class Based React Component
+class Modal extends Component {
+  constructor(props) {
+    super(props);
+    console.log(props);
```
ZakkMan: Can you remove any `console.log` statements?

client/src/components/Modal/Modal.js
```diff
@@ -0,0 +1,43 @@
...
+
+    // Default CSS class to apply to the Component
+    this.state = {
+      classList: "Modal",
+    };
+  }
```
ZakkMan: Please just put the class name in the `className` attribute

client/src/components/Modal/Modal.js
```diff
@@ -0,0 +1,43 @@
+import React, { Component } from 'react';
+import "./Modal.css";
+import "../../App.css";
+
+
+// Class Based React Component
+class Modal extends Component {
```
ZakkMan: I think this could a function component

client/src/components/Modal/Modal.js
```diff
@@ -0,0 +1,43 @@
...
+    super(props);
+    console.log(props);
+
+    // Default CSS class to apply to the Component
+    this.state = {
+      classList: "Modal",
+    };
+  }
+  // Runs after Component is loaded in the broswer
+  componentDidMount() { }
```
ZakkMan: Can you delete this, since you're not using it

client/src/components/Modal/Modal.stories.js
```diff
@@ -0,0 +1,39 @@
...
+  alert('Modal Closed!')
+}
+
+storiesOf('Modal', module)
+  .add("Follow Modal", () => 
+    <Modal message={"Are you sure you want to follow"} 
+          confirmText={"Confirm"} 
+          cancelText={"Cancel"}
+          onConfirm={handleConfirm} 
+          onCancel={handleCancel}
```
ZakkMan: I think `onCancel` and `onClose` are the same - we should invoke the same callback when they click Cancel as when they click the X

client/src/components/Modal/Modal.css
ZakkMan: I think this needs a little more padding around it

Screen Shot 2019-12-11 at 10 41 15 AM

client/src/components/Modal/Modal.stories.js
```diff
@@ -0,0 +1,39 @@
...
+function handleCancel (evt){
+  evt.preventDefault()
+  alert('Canceled!')
+}
+function handleClose (evt){
+  evt.preventDefault()
+  alert('Modal Closed!')
+}
+
+storiesOf('Modal', module)
```
ZakkMan: Can you work through and create a story for having the modal pop up and close out? Like just a button or link that we can click that triggers the open and then with close functions

caioingber:
A bit confused on what still needs to be worked on for this PR. Any ideas?

caioingber:
A bit confused on what still needs to be worked on for this PR - seems like it is functional and responsive in storybook. Any ideas?

client/src/components/FunctionalModal/FunctionalModal.css
```diff
@@ -0,0 +1,29 @@
+.FunctionalModal{
```
ZakkMan: Can you make the class names consistent with the CSS naming convention?

client/src/components/Modal/Modal.stories.js
caioingber: Hey @ZakkMan from what I can tell, the `FunctionalModal` component is the one that demonstrates functionality as it includes a button and the Modal component itself, and it is where state lives so that showModal prop can be toggle true/false

caioingber:
modal was changed to a functional component, updated styling names to match convention. See comment above in regards to FunctionalModal vs. Modal

caioingber:
Set up a ModalContainer.js file in the Modal directory - per our convo, this contains both the functional Modal component as well as a Button component which allows the user to open the modal. Updated stories to utilize the ModalContainer component to demonstrate functionality. If I understood correctly, this is what you were getting at. Leaving the FunctionalModal component for now, in case there are issues with my updates.

client/src/components/Modal/Modal.css
```diff
@@ -0,0 +1,125 @@
...
+  text-decoration: none;
+  color: black;
+}
+
+.Modal__body {
+  width: 100%;
+  height: 60%;
+}
+
+.Modal__messageBox {
```
ZakkMan: With the naming convention we're using, I think this should be `.Modal__message-box`

ZakkMan: (Sorry for not catching this earlier 😳 )

client/src/components/Modal/Modal.css
```diff
@@ -0,0 +1,125 @@
...
+  height: 23%;
+  display: flex;
+  flex-direction: row;
+}
+
+.Modal__footerButtons {
+  width: 100%;
+}
+
+.Modal__footerButtonConfirm {
```
ZakkMan: Does the button component not work here? It's okay if it doesn't but that would be ideal

client/src/components/Modal/Modal.js
```diff
@@ -0,0 +1,39 @@
...
+
+const Modal = props => {
+  return (
+    <div
+      className="Modal"
+      style={{
+        transform: props.showModal ? "translateY(0vh)" : "translateY(-100vh)",
+        opacity: props.showModal ? "1" : "0"
+      }}
+    >
```
ZakkMan: This is a good way of handling this. Another is to create 'state' classes like `.is-visible` or `.is-hidden` and then toggle adding or removing those classes.

ZakkMan: You don't have to go with that

client/src/components/Modal/ModalContainer.js
```diff
@@ -0,0 +1,54 @@
+import React, { Component } from "react";
```
ZakkMan: Do you mind putting an underscore in the beginning of the file name here? I forgot to mention that when we spoke on slack. That's a way of signalling to other developers that this is a "hidden" file/class/whatever. That will make it extra clear that people shouldn't use this component

ZakkMan: Just to the file name

client/src/components/Modal/ModalContainer.js
```diff
@@ -0,0 +1,54 @@
...
+  };
+
+  handleOpenModal = () => {
+    this.setState({
+      modalOpen: true
+    });
+  };
+
+  render() {
+    console.log(this.state.modalOpen);
```
ZakkMan: Can you remove this `console.log` - we never want to push that to prod

client/src/components/Navbar/Navbar.js
```diff
@@ -15,13 +15,13 @@ class Navbar extends Component {
     this.state = {
```
ZakkMan: Do we need these changes? They seem unrelated to what you're working on? Can you maybe reset this file so these changes aren't part of this PR?

client/src/index.js
```diff
@@ -5,6 +5,7 @@ import App from "./App";
 import * as serviceWorker from "./serviceWorker";
 import { BrowserRouter as Router } from "react-router-dom";
 
+
```
ZakkMan: Same with this change - can you reset it so its not in this PR?

client/src/components/Modal/Modal.css
caioingber: From what I can tell, the styling for these buttons are quite specific (different color, borders, sizes) from what is prescribed in the storybook buttons, and they appear to be standalone

client/src/components/Modal/Modal.js
caioingber: Yea, whoever in 33 was working on it, implemented that functionality, it's kinda nice since it allows for a nice sliding transition

caioingber:
This should be fixed, the changes from app, index and navbar all came from merging develop locally, so there shouldn't be any changes to the develop branch. Confirmed storybook path is correct Layout/Modal. Resolved previous conversations about naming conventions, etc

PR #122, opened 2019-12-08
Component pink header

cenwachukwu:
Created 5 components for the pink banner


ZakkMan:
@cenwachukwu Make sure you assign yourself the PR

client/src/components/PinkHeader/PinkHeader.css
```diff
@@ -0,0 +1,61 @@
...
+    height: 48px;
+    display: flex;
+    flex-direction: row;
+}
+
+
+.container-1 input#search {
+    width: 356px;
+    height: 48px;
+    background-color: #FAFAFA;
```
wchrapcynski: Let's use the css variables for colors here. For instance, that background color would be:
background-color: var(--white);
You can refer to App.css for any variables.
Keep rockin!

vmipfdr:
@ZakkMan this might just need to be closed @Underwaterr

client/src/components/PinkHeader/PinkHeader.css
```diff
@@ -0,0 +1,63 @@
+*{
+    font-family: var(--default-font)
+}
```
ZakkMan: Can you delete this? if we need to set a style globally, we should do so in `index.css`. I don't think we want to set this globally though

client/src/components/PinkHeader/PinkHeader.css
```diff
@@ -0,0 +1,63 @@
...
+
+.PinkHeader {
```
ZakkMan: Can you come up with a more abstract name? Not all the headers are pink

client/src/components/PinkHeader/PinkHeader.css
```diff
@@ -0,0 +1,63 @@
...
+    background-repeat:no-repeat;
+    background-size: cover;
+    width: 1440px;
+    height: 231px;
+    text-align: center;
+    padding-top: 3px;
+    box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
+}
+
+h3{
```
ZakkMan: Can you nest this following the class naming convention described in the readme?

client/src/components/PinkHeader/PinkHeader.css
```diff
@@ -0,0 +1,63 @@
...
+    font-family: 'Reem Kufi', sans-serif;
```
ZakkMan: Can you use the css variable for this?

client/src/components/PinkHeader/PinkHeader.css
```diff
@@ -0,0 +1,63 @@
...
+    color: #FAFAFA;
```
ZakkMan: Can you use the CSS variable for all colors too

client/src/components/PinkHeader/PinkHeader.css
```diff
@@ -0,0 +1,63 @@
+*{
+    font-family: var(--default-font)
+}
+
+.PinkHeader {
+    background-image: var(--profile-banner);
+    background-repeat:no-repeat;
+    background-size: cover;
+    width: 1440px;
```
ZakkMan: Why does this have an explicit width? We'll want it to take up the width of the page, which may be more or less than 1440px

client/src/components/PinkHeader/PinkHeader.css
```diff
@@ -0,0 +1,63 @@
...
+    height: 231px;
```
ZakkMan: Does this need an explicit height? Is there a way we can add a consistent amount of padding? If the content in the header is too big, this will break it

client/src/components/PinkHeader/PinkHeader.css
```diff
@@ -0,0 +1,63 @@
...
+    text-align: center;
+    padding-top: 3px;
+    box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
+}
+
+h3{
+    font-family: 'Reem Kufi', sans-serif;
+    color: #FAFAFA;
+    font-size: 48px;
```
ZakkMan: Are there CSS variables for font sizes?

client/src/components/PinkHeader/PinkHeader.css
```diff
@@ -0,0 +1,63 @@
...
+    line-height: 72px;
+    text-align: center;
+    margin: 25px 0 0 0;
+}
+
+.box2{
```
ZakkMan: Can you find a better name for this class? Again using the naming convention in the readme?

client/src/components/PinkHeader/PinkHeader.css
```diff
@@ -0,0 +1,63 @@
...
+    font-size: 48px;
+    line-height: 72px;
+    text-align: center;
+    margin: 25px 0 0 0;
+}
+
+.box2{   
+    margin: 0 auto;
+    width: 356px;
+    height: 48px;
```
ZakkMan: Do we need explicit height and width here? That may make responsive difficult

client/src/components/PinkHeader/PinkHeader.css
```diff
@@ -0,0 +1,63 @@
...
+    display: flex;
+    flex-direction: row;
+}
+
+
+.container-1 input#search {
```
ZakkMan: Can you find a better name and selector here

ZakkMan: We probably shouldn't be using an ID here

client/src/components/PinkHeader/PinkHeader.css
```diff
@@ -0,0 +1,63 @@
...
+    width: 356px;
+    height: 48px;
```
ZakkMan: Same comment re height and width

client/src/components/PinkHeader/PinkHeader.css
ZakkMan: +1

client/src/components/PinkHeader/PinkHeader.css
```diff
@@ -0,0 +1,63 @@
...
+    background-color: #FAFAFA;
+    border: 1px solid #E5E5E5;
+    font-style: normal;
+    font-weight: normal;
+    font-size: 24px;
+    line-height: 36px;
```
ZakkMan: I think there are CSS variables for this stuff

client/src/components/PinkHeader/PinkHeader.css
```diff
@@ -0,0 +1,63 @@
...
+    padding: 15px 0 10px 0;
+    font-family: 'Reem Kufi', sans-serif;
```
ZakkMan: Please use the CSS variable

client/src/components/PinkHeader/PinkHeader.css
```diff
@@ -0,0 +1,63 @@
...
+    color:  rgba(37, 37, 37, 0.3);
```
ZakkMan: CSS variable here too

client/src/components/PinkHeader/PinkHeader.css
```diff
@@ -0,0 +1,63 @@
...
+    padding-left: 65px;
+    border-radius: 3px;
+    box-shadow: 0px 4px 4px rgba(0, 0, 0, 0.25);
+}
+
+.container-1 .icon{
```
ZakkMan: Can you find a better class name here?

client/src/components/PinkHeader/PinkHeader.css
```diff
@@ -0,0 +1,63 @@
...
+    position: absolute;
+    top: 47%;
+    margin-left: 25px;
+    margin-top: 14px;
+    z-index: 1;
```
ZakkMan: Does this need to be positioned absolutely? Is there a better way to do this?

client/src/components/PinkHeader/PinkHeader.css
```diff
@@ -0,0 +1,63 @@
...
+}
+
+img{
```
ZakkMan: Can you provide a class name here? Otherwise these stills will clash with other image styling in the app

client/src/components/PinkHeader/PinkHeader.js
```diff
@@ -0,0 +1,102 @@
...
+class Banner extends Component {
+  constructor(props) {
+    super(props);
+    console.log(props);
+
+    // Default CSS class to apply to the Component
+    this.state = {
+      classList: "PinkHeader"
+    };
+  }
```
ZakkMan: Since there's no state in the app, can you make this a function component?

client/src/components/PinkHeader/PinkHeader.js
```diff
@@ -0,0 +1,102 @@
...
+
+  render() {
```
ZakkMan: Can you find a way to write this that doesn't include so much duplication? We shouldn't need these massive if statements

client/src/components/PinkHeader/PinkHeader.stories.js
```diff
@@ -0,0 +1,29 @@
...
+    <PinkHeader
+        Trip/>
+  ))
+  .add("Directory: My Friends", () => (
+    <PinkHeader
+        Friends
+        placeholder="Search My Directory"/>
+  ))
+  .add("Directory: My Groups", () => (
+    <PinkHeader
```
ZakkMan: This component should probably just be called Header

client/src/components/PinkHeader/PinkHeader.stories.js
```diff
@@ -0,0 +1,29 @@
...
+        bookTrip
+        placeholder="Search Trips"/>
+  ))
+  .add("Trip", () => (
+    <PinkHeader
+        Trip/>
+  ))
+  .add("Directory: My Friends", () => (
+    <PinkHeader
+        Friends
```
ZakkMan: Rather than thinking about this in terms of types of header (Trip Header, Friends Header, Groups Header, etc) think of this in terms of a modular header. We want all those combinations to work together

client/src/components/PinkHeader/PinkHeader.css
vmipfdr: removed from `PInkHeader.css` and moved to `index.css` to see if we want to run it, else I'll comment it out.

client/src/components/PinkHeader/PinkHeader.css
vmipfdr: Yeup

client/src/components/PinkHeader/PinkHeader.css
vmipfdr: yes

client/src/components/PinkHeader/PinkHeader.css
vmipfdr: Yes

client/src/components/PinkHeader/PinkHeader.css
vmipfdr: yes

client/src/components/PinkHeader/PinkHeader.css
vmipfdr: done

vmipfdr:
@ZakkMan I looked at everything in this pull request vs what is already in the Banner component and it's duplicate work.  I reviewed every component in the Banner component and we have the functionality.  I recommend we close this Pull Request.

Issue #60, opened 2019-12-06
Component: Previous Trips/Hashtags Card


TylerEikenberg:
<img width="520" alt="Screen Shot 2019-12-06 at 11 55 21 AM" src="https://user-images.githubusercontent.com/55251251/70348509-d0468200-1830-11ea-8d93-f05e95b622d6.png">
<img width="369" alt="Screen Shot 2019-12-06 at 11 33 40 AM" src="https://user-images.githubusercontent.com/55251251/70348515-d2a8dc00-1830-11ea-9d9d-578832e1399a.png">

**Props**
- _Previous Trip_ or _Hashtag_ to determine which version is displayed.
- **IF props === Previous Trip**  
    -Props for array of names of previous trip locations
- **IF props === Hashtag**  
    -Props for array of hashtags (max length of 10)

Styles: 
- ```Height```: **265px**
- ```Width```: **505px**

- ```Background-Color```: **6A4C93**

bmillsmc:
This component has been split into two separate issues at #171 and #172.

PR #123, opened 2019-12-08
Component: Previous Trips/Hashtag Card PR

TylerEikenberg:
Issue #60 . This component renders different versions with the prop ```type``` of ```hashtags``` and ```trips```. Pass in an array of trips or hashtags using the prop ```data```. If the trips data array length is greater than 10 it is automatically trimmed down to only use the last 10 items.


TylerEikenberg:
@ZakkMan I took your suggestions and significantly revamped this component. Now theres much less code, all the text content (besides the data) is passed in as children.

client/src/components/PreviousTripHashtagCard/PreviousTripHashtagCard.js
```diff
@@ -0,0 +1,36 @@
+import React from "react";
+import "./PreviousTripHashtagCard.css";
+import "../../App.css";
+
+// Function based React Component
+const PreviousTripHashtagCard = props => {
+  console.log(props);
```
ZakkMan: Remove and `console.log`s!

client/src/components/PreviousTripHashtagCard/PreviousTripHashtagCard.js
```diff
@@ -0,0 +1,36 @@
...
+  // Default Class to apply to Component
+  const classList = `prev-hash-card-container`;
```
ZakkMan: If you're not going to change or update this class, then just add it as a string to the `className` property?

client/src/components/PreviousTripHashtagCard/PreviousTripHashtagCard.js
```diff
@@ -0,0 +1,36 @@
+import React from "react";
+import "./PreviousTripHashtagCard.css";
+import "../../App.css";
+
+// Function based React Component
+const PreviousTripHashtagCard = props => {
```
ZakkMan: Can you think of a more general, semantic name for this?

client/src/components/PreviousTripHashtagCard/PreviousTripHashtagCard.js
```diff
@@ -0,0 +1,36 @@
...
+    data = props.data.slice(dataLength - 10, dataLength);
+  } else {
+    data = props.data;
+  }
+
+  return (
+    <div className={classList}>
+      <div className="trips-content-container">
+        {props.children}
+        {props.type === "trips slider" ? <div className="trip-slider-arrow" /> : null}
```
ZakkMan: Maybe slider could be a child? Or we could have a `SliderCard` component? This feels like too many things in a single component

client/src/components/PreviousTripHashtagCard/PreviousTripHashtagCard.stories.js
```diff
@@ -0,0 +1,76 @@
...
+  .add("Previous Trips, No Data", () => (
+    <PreviousTripHashtagCard data={[]} type="trips">
+      <h1>Previous Trips</h1>
+      <h2>None with CoTripper yet!</h2>
+      <a href="#">Book a Trip Now!</a>
+    </PreviousTripHashtagCard>
+  ))
+  .add("Previous Trips", () => (
+    <PreviousTripHashtagCard
+      data={["New York", "Mexico", "Wakanda", "Hawaii", "Japan", "Canada", "Italy"]}
```
ZakkMan: One thing to note is that these hashtags might be links. So the data that would come through here as a prop might look more like:

```js
data={[{link: '/new-york', title: 'New York'},{link: '/mexico', title: 'Mexico'},{link: '/wakanda', title: 'Wakanda'},]}
```

ZakkMan: You can maybe find a way to handle both, or have this list be a `ul` that the user passes in as a child (like the text)

Issue #20, opened 2019-12-05
Component: Input Type Select


tissakhosla:
![image](https://user-images.githubusercontent.com/53238865/70267922-703ad780-176d-11ea-8e62-32e1d3a637bc.png)

PR #124, opened 2019-12-08
Component: Input Type Select PR

torykling:
#20 

Built a select input component that takes an array of options and a callback function. 

I wasn't sure about how to set up the functionality of this component. Should the callback function take this.state.value as a prop?


torykling:
@ZakkMan I deleted the unused lifecycle methods, made options an array of objects instead of strings, and passed the target value as a prop to the onSelect method. 

Remaining questions:
I added proptypes and made onSelect a required prop because I kept getting error messages when I tried to set a method as a default onSelect prop. Is there a better way to handle this? 

Does the rest of the logic make sense?

As for styling, I tweaked it a bit but not much.  I know the font isn't right but I listed it as Reem Kufi and was hoping we'd import that font on the index.html page in the public folder, which I don't see here. Is there another way to link fonts?  Is there something else I should do?

client/src/components/InputSelect/InputSelect.css
```diff
@@ -0,0 +1,18 @@
+.InputSelect {
+  width: 20rem;
+  height: 3rem;
+  background: #fafafa;
+  border: 2px solid #e5e5e5;
```
ZakkMan: For colors, if you update your branch with `develop`, you should get the CSS variables that Bill created. Do you mind using those for the color and font properties?

client/src/components/InputSelect/InputSelect.js
```diff
@@ -0,0 +1,55 @@
+import React, { Component } from "react";
+import "./InputSelect.css";
+import PropTypes from "prop-types";
+import "../../App.css";
```
ZakkMan: You shouldn't need to do this anymore - Lorenzo made a PR that set this for storybook globally

client/src/components/InputSelect/InputSelect.js
```diff
@@ -0,0 +1,55 @@
...
+
+// Class Based React Component
+class InputSelect extends Component {
+  constructor(props) {
+    super(props);
+    console.log(props);
```
ZakkMan: Can you delete any `console.log` statements?

client/src/components/InputSelect/InputSelect.js
```diff
@@ -0,0 +1,55 @@
...
+      classList: "InputSelect",
+      value: "select"
+    };
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  handleChange(e) {
+    e.preventDefault();
+    this.setState({ value: e.target.value });
+    this.props.onSelect(e.target.value);
```
ZakkMan: We probably want this to be synchronous, which you like this:

```js
this.setState({ value: e.target.value }, () => this.props.onSelect(e.target.value));
```

client/src/components/InputSelect/InputSelect.js
```diff
@@ -0,0 +1,55 @@
...
+  options: [
+    { value: "option1", title: "Option 1" },
+    { value: "option2", title: "Option 2" }
+  ]
+};
+InputSelect.propTypes = {
+  optionPrefix: PropTypes.string,
+  options: PropTypes.array,
+  onSelect: PropTypes.func.isRequired
+};
```
ZakkMan: 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 
This is so dope!

client/src/components/InputSelect/InputSelect.js
```diff
@@ -0,0 +1,52 @@
...
+import "./InputSelect.css";
+import PropTypes from "prop-types";
+
+// Class Based React Component
+class InputSelect extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      classList: "InputSelect",
```
ZakkMan: If you just have this one class, then you can put the string in your JSX for the `className` property

ZakkMan:
@torykling Can you update your branch with `develop` and resolve any merge conflicts

Issue #87, opened 2019-12-06
Component: Forum Comment 


cjohnston96:
<img width="620" alt="Screen Shot 2019-12-06 at 2 43 31 PM" src="https://user-images.githubusercontent.com/30026036/70351390-d4759e00-1836-11ea-8a38-971a117a3e59.png">

**(Molecular)**
- User Profile Pic/Name **(Component)**
- Forum Comment Body **(component)**
- Likes/replies **(props)**

reisrael4:
This was moved over to a component just called "Comment" by @colleenobrien3.

Issue #85, opened 2019-12-06
Component: Comment Body


cjohnston96:
<img width="542" alt="Screen Shot 2019-12-06 at 2 40 24 PM" src="https://user-images.githubusercontent.com/30026036/70351182-63ce8180-1836-11ea-9398-ce29eb213463.png">

**(Atomic)**
- Comment Body **(props)**
- Date **(props)**

PR #128, opened 2019-12-08
Component: Comment Body PR

cjohnston96:
Issue #85

I added no padding to the left or right so that when this is used within issue #87, there is more maneuverability with this atomic component. I wasn't exactly sure how I should use props for the comment body or date at this point, so i just added in where I thought the code should go within the component. It will take in the body and date as props for each comment, and right now the displayed text is just an example of how it should look.


client/src/components/CommentBody/CommentBody.js
```diff
@@ -0,0 +1,31 @@
...
+    <div className={classList}>
+      <p className="Body">
+      Sed ut perspiciatis unde omnis iste natus error sit voluptatem 
+      accusantium doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo 
+      inventore v eritatis et quasi architecto beatae vitae dicta sunt explicabo. Nemo enim 
+      ipsam voluptatem quia voluptas sit aspernatur aut odit
+      {/* {this.props.comment} */}
+      </p>
+      <p className="Date">
+      Date <br/>
```
ZakkMan: Don't use `<br /> - use CSS to accomplish this

client/src/components/CommentBody/CommentBody.js
```diff
@@ -0,0 +1,31 @@
...
+
+
+  return(
+    <div className={classList}>
+      <p className="Body">
+      Sed ut perspiciatis unde omnis iste natus error sit voluptatem 
+      accusantium doloremque laudantium, totam rem aperiam, eaque ipsa quae ab illo 
+      inventore v eritatis et quasi architecto beatae vitae dicta sunt explicabo. Nemo enim 
+      ipsam voluptatem quia voluptas sit aspernatur aut odit
+      {/* {this.props.comment} */}
```
ZakkMan: Can you swap in these props for the hardcoded text?

client/src/components/CommentBody/CommentBody.js
```diff
@@ -0,0 +1,31 @@
...
+      </p>
+      <p className="Date">
+      Date <br/>
+      Time
+      {/* {this.props.date} */}
```
ZakkMan: Can you swap in the prop here too?

client/src/components/CommentBody/CommentBody.js
```diff
@@ -0,0 +1,31 @@
+
+import React from 'react';
+import "./CommentBody.css";
+
+// Function based React Component
+const CommentBody = (props) => {
+  // Default Class to apply to Component
+  // No props right now but when there is data we can use 
+  // this.props.comment and this.props.date
+  let classList = `CommentBody`;
```
ZakkMan: If you're not going to be updating/editing the class list, then just drop this into the JSX for the `className` property

client/src/components/CommentBody/CommentBody.css
```diff
@@ -0,0 +1,16 @@
...
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+    align-items: flex-start;
+    width: 737px;
+    height: auto;
+    background-color: rgb(196, 196, 196);
+    border: 1px solid rgb(171, 171, 171);
+    font-size: 14px;
+    font-family: "Open Sans";
```
ZakkMan: Can you use the CSS Variables for the font properties?

client/src/components/CommentBody/CommentBody.css
```diff
@@ -0,0 +1,16 @@
+
+.CommentBody {
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+    align-items: flex-start;
+    width: 737px;
+    height: auto;
+    background-color: rgb(196, 196, 196);
+    border: 1px solid rgb(171, 171, 171);
```
ZakkMan: Can you use the CSS variables for the colors?

client/src/components/CommentBody/CommentBody.css
```diff
@@ -0,0 +1,16 @@
+
+.CommentBody {
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+    align-items: flex-start;
+    width: 737px;
```
ZakkMan: This probably shouldn't have an explicit width like this - we'll want it to be responsive!

client/src/components/CommentBody/CommentBody.css
```diff
@@ -0,0 +1,21 @@
+@import url(http://fonts.googleapis.com/css?family=Open+Sans);
+@import "../../App.css";
```
ZakkMan: These are set globally, so you shouldn't need to do this

client/src/components/CommentBody/CommentBody.css
```diff
@@ -0,0 +1,21 @@
...
+.CommentBody {
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+    align-items: flex-start;
+    width: 60%;
+    height: auto;
+    background-color: var(--gray-dark);
+    font-size: 14px;
+    font-family: 'Open Sans', sans-serif;
```
ZakkMan: Can you use the CSS var for the font?

client/src/components/CommentBody/CommentBody.js
```diff
@@ -0,0 +1,32 @@
+
+import React from 'react';
+import "../../App.css"
```
ZakkMan: You shouldn't need to import `App.css` here either

client/src/components/CommentBody/CommentBody.js
```diff
@@ -0,0 +1,32 @@
...
+
+// Function based React Component
+const CommentBody = (props) => {
+  // Default Class to apply to Component
+  // No props right now but when there is data we can use 
+  // this.props.comment and this.props.date
+  let classList = `CommentBody`;
+  const body = props.bodyProps;
+  const date = props.dateProps;
+  const time = props.timeProps;
```
ZakkMan: Can you use destructing here to clean this up?

Also, you don't need to put `Prop` in the name of a prop (i.e. just `body`, not `bodyProp`)

client/src/components/CommentBody/CommentBody.js
```diff
@@ -0,0 +1,32 @@
...
+
+  return(
+    <div className={classList}>
+      <p className="Body">
+        {body}
+      </p>
+      <div className="Date">
+        <p>
+          {date}
+          <br/>
```
ZakkMan: Please don't use the `<br />` element - instead, use CSS to style this

client/src/components/CommentBody/CommentBody.js
ZakkMan: @cjohnston96 don't forget this one

Issue #39, opened 2019-12-06
Component: Info Card Issue


llucas314:
<img width="193" alt="Screen Shot 2019-12-05 at 9 17 45 PM" src="https://user-images.githubusercontent.com/42591694/70290075-c8d99700-17a4-11ea-9aa3-83c085c9cb1d.png">

Info Card Component:

Properties:

* `group`: {
    `name`: string,
    `members`: int,
    `location`: string,
    `state`: string
}

* `trip`: {
    `location`: string,
    `month`: string,
    `year`: int
}

If the `group` object is passed as a prop, display the objects with the following css:

- `width`: 220px
- `height`: 150px
- `border-radius`: 3px
- `background-color`: #FB836F
- `filter`: drop-shadow(0px 4px 4px rgba(0, 0, 0, 0.25));

`name`: 

- Font Family: Reem Kufi (https://fonts.google.com/specimen/Reem+Kufi)
- Font Style: Regular
- Font Size: 24px
- Line Height: 36px
- Line Height: 100%
- Align: Left
- Vertical Align: Top
- Color: #FFFFFF

`members`:

- Font Family: Reem Kufi (https://fonts.google.com/specimen/Reem+Kufi)
- Font Style: Regular
- Font Size: 14px
- Line Height: 21px
- Line Height: 100%
- Align: Left
- Vertical Align: Top
- Color: #FFFFFF

`location` & `state`:

- Font Family: Reem Kufi (https://fonts.google.com/specimen/Reem+Kufi)
- Font Style: Regular
- Font Size: 14px
- Line Height: 21px
- Line Height: 100%
- Align: Right
- Vertical Align: Top
- Color: #FFFFFF

If the `trip` object is passed as a prop, display the objects with the following css:

- `width`: 218px
- `height`: 145px
- `border-radius`: 3px
- `background-color`: rgba(196, 196, 196, 0.25)

`location`:

- Font Family: Reem Kufi (https://fonts.google.com/specimen/Reem+Kufi)
- Font Style: Regular
- Font Size: 24px
- Line Height: 36px
- Line Height: 100%
- Align: Center
- Vertical Align: Top
- Color: #FFFFFF

`month` & `year`:

- Font Family: Reem Kufi (https://fonts.google.com/specimen/Reem+Kufi)
- Font Style: Regular
- Font Size: 18px
- Line Height: 27px
- Line Height: 100%
- Align: Center
- Vertical Align: Top
- Color: #FAFAFA

[...]

PR #129, opened 2019-12-08
Component: TripCard PR

bmillsmc:
This is a pull request for the Image Cards talked about in issue #172.

The background image is controlled by a `url` prop, but keep in mind this can be a file path as well. Text should be passed in using an array through the `text` prop.


bmillsmc:
Just realized this is a subset of a component in #39 which Shelby is currently on. I'll start on something else.

client/src/components/ImageCard/ImageCard.css
```diff
@@ -0,0 +1,35 @@
...
+  width: 100%;
+  display: flex;
+  flex-direction: column;
+  height: 70%;
+  align-items: center;
+}
+
+.textHolder h3 {
+  color: rgb(242, 235, 235);
+  font: 1.2vw Helvetica, Sans-Serif;
```
ZakkMan: Hey, can you use the css variables for font and color properties?

client/src/components/ImageCard/ImageCard.css
```diff
@@ -0,0 +1,35 @@
...
+.textHolder {
+  position: absolute;
+  top: 12%;
+  left: 0;
+  width: 100%;
+  display: flex;
+  flex-direction: column;
+  height: 70%;
+  align-items: center;
+}
```
ZakkMan: Hey, do we need to use absolute positioning here? Could we apply the image as a background image and then just align the text with flexbox

client/src/components/ImageCard/ImageCard.js
```diff
@@ -0,0 +1,31 @@
...
+          {text}
+        </h3>
+      );
+    }
+    return (
+      <h3 className="textEl" key={index}>
+        {text}
+      </h3>
+    );
+  });
```
ZakkMan: Can you cut this down? Do you need this conditional? If so, could you just get away with a ternary?

client/src/components/ImageCard/ImageCard.js
```diff
@@ -0,0 +1,31 @@
...
+  return (
+    <div className={classList}>
+      {image}
+      <div className="textHolder">{textEls}</div>
```
ZakkMan: Can you use a more semantic name than `textEls`?

client/src/components/ImageCard/ImageCard.js
```diff
@@ -0,0 +1,31 @@
...
+      );
+    }
+    return (
+      <h3 className="textEl" key={index}>
+        {text}
+      </h3>
+    );
+  });
+  return (
+    <div className={classList}>
```
ZakkMan: If you're not changing or updating this class, can you just reference the class as a string here, instead of using this variable?

client/src/components/ImageCard/ImageCard.stories.js
```diff
@@ -0,0 +1,7 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import ImageCard from "./ImageCard";
+
+storiesOf("ImageCard", module).add("Default", () => (
+  <ImageCard url="https://wallpaperaccess.com/full/144067.jpg" text={["Hawaii", "May 2020"]} />
```
ZakkMan: Could we break up the `text` prop into separate props? Maybe something like `bodyText` and `footerText`?

bmillsmc:
#175 is the new issue that this pull request relates to

client/src/components/TripCard/TripCard.css
```diff
@@ -0,0 +1,30 @@
+.TripCard__date,
+.TripCard__location {
+  font-family: "Reem Kufi";
```
ZakkMan: Can you use the CSS variables for any font and color settings?

client/src/components/TripCard/TripCard.css
```diff
@@ -0,0 +1,30 @@
...
+  margin: 14% 0 0;
+}
+
+.TripCard {
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-end;
+  width: 100%;
+  height: 100%;
+}
```
ZakkMan: Put the root class at the top of the CSS file

client/src/components/TripCard/TripCard.stories.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import TripCard from "./TripCard";
+import example1 from "../../../../assets/images/card_small2.png";
+
+storiesOf("TripCard", module)
+  .add("Default", () => (
+    <TripCard
+      path="https://wallpaperaccess.com/full/144067.jpg"
```
ZakkMan: Let's call the image prop `src`

client/src/components/TripCard/TripCard.css
```diff
@@ -0,0 +1,30 @@
+.TripCard {
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-end;
+  width: 100%;
+  height: 100%;
```
erinkelley27: Change from percentage to rem. Talk to @shelbyvjacobs if you have any questions!

bmillsmc: I fixed all the percentages except the ones you highlighted, that div needs to be 100% of its parent, so the bug @shelbyvjacobs was talking about won't affect it.

Issue #73, opened 2019-12-06
Component: Post Body


cjohnston96:
<img width="659" alt="Screen Shot 2019-12-06 at 12 17 56 PM" src="https://user-images.githubusercontent.com/30026036/70349602-04bb3d80-1833-11ea-9226-b24df356663c.png">

**(Molecular)**
- Title
- Date
- Tags **(Props)**
- Post body

PR #133, opened 2019-12-09
Component: Post Body PR

torykling:
#73

The rough styling is there but it's hard to tell how close it is with the fonts being different. Tomorrow I'll figure out how to get the fonts imported correctly. 

Question about css class names - If I name a class name for this component .title or .date could the styles I apply affect anyone else's component with the same class name? I made all these class names post-body- something but I'll change it if they won't conflict. 

I wasn't sure how the data would look here for the props, so I guessed.


client/src/components/PostBody/PostBody.css
```diff
@@ -0,0 +1,49 @@
+.PostBody {
+  background: #fafafa;
+  width: 50rem;
+  font-family: "Open Sans", sans-serif;
+  color: #252525;
```
ZakkMan: Can you please use the CSS variables for color and font properties?

client/src/components/PostBody/PostBody.css
```diff
@@ -0,0 +1,49 @@
...
+}
+
+.post-body-top-right {
+  display: flex;
+  flex-direction: column;
+  align-items: flex-end;
+}
+
+.post-body-title {
+  font-family: "Reem Kufi", sans-serif;
```
ZakkMan: Can you use the CSS variable for the font name?

client/src/components/PostBody/PostBody.js
```diff
@@ -0,0 +1,41 @@
+import React from "react";
+import "./PostBody.css";
+
+// Function based React Component
+const PostBody = props => {
+  // Default Class to apply to Component
+  let classList = `PostBody`;
```
ZakkMan: If you're not updating or changing the class name, then you can just reference the class as a string in the JSX and remove this extra variable

client/src/components/PostBody/PostBody.css
torykling: Sounds good. I asked Bill to add Open Sans to the app.css. Once his PR gets merged, I can add that variable here.

client/src/components/PostBody/PostBody.js
```diff
@@ -0,0 +1,39 @@
+import React from "react";
+import "./PostBody.css";
+import "../../App.css";
```
ZakkMan: You shouldn't need to import the `App.css` file anymore - it's been added to the storybook global config

client/src/components/PostBody/PostBody.css
```diff
@@ -0,0 +1,49 @@
+.PostBody {
+  background: var(--white);
+  width: 50rem;
+  font-family: var(--paragraph-font);
+  color: var(--black);
+}
```
ZakkMan: I think you need some padding here

client/src/components/PostBody/PostBody.js
torykling: Removed this. Bill's going to import the bold version of open sans.

Issue #137, opened 2019-12-09
Add in Open Sans font & gray Issue


wchrapcynski:
Open Sans is needed for paragraphs.

wchrapcynski:
Added in a link to Open Sans on Google fonts. Also cleaned up default React code in App.js and App.css.

wchrapcynski:
Another thing missing are some shades of gray. I added those in to App.css.

PR #139, opened 2019-12-09
Css variables - Adding Open Sans font PR

wchrapcynski:
Per issue #137, Open Sans has been added in to App.css.


wchrapcynski:
Also looks like there were a couple shades of gray missing. Added those into App.css.

client/src/App.js
```diff
@@ -1,24 +1,13 @@
...
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
+<<<<<<< HEAD
+=======
+  
+>>>>>>> parent of 08b33aa... Revert "Updated css variables with another font that was needed. Also cleaned up App and removed default React setup code."
     </div>
```
ZakkMan: Can you fix this piece

wchrapcynski:
@ZakkMan This should be good to go now.

client/src/App.js
wchrapcynski: I accidentally edited in the wrong branch but I should have reverted back. Hopefully all should be good now.

Issue #59, opened 2019-12-06
Components: Nav Link & Nav Link Dropdown Menu


DallasLind:
NOTE: images not to scale

![image](https://user-images.githubusercontent.com/54913865/70455246-f0b54d00-1a79-11ea-8af5-69b91fd633cf.png)
![image](https://user-images.githubusercontent.com/54752666/70348685-37643680-1831-11ea-8375-75f0e42f0a59.png)

bhearn7:
NEED to fix Redirect once routes are made
NEED to fix any styling issues once this is inserted into the NavBar

PR #143, opened 2019-12-09
completed atomic nav bar link components

bhearn7:
Per issue #59 

created NavBarLink component that takes 3 props:
- text (i.e. My Directory, Community, Forum, Book A Trip)
- to (route for when the user clicks the nav bar link)
- menuList (Array of objects for the dropdown menu)

created NavBarDropdownMenu component that takes 1 prop from NavBarLink:
- menuList (Array of objects)
--- text (i.e. Explore People, Join A Group)
--- to (route for when a user clicks the dropdown menu item)


client/src/components/NavLink/NavLink.css
```diff
@@ -0,0 +1,22 @@
...
+  top: -500%;
+  left: 0;
+  width: 100%;
+}
+.dropdown-show {
+  position: absolute;
+  top: 100%;
+  left: 0;
+  width: 100%;
+}
```
ZakkMan: This is a good approach - I like you're thinking. A better way to do this is to have the dropdown in position and then hide and show it by setting `opacity` from `0` to `1` and changing the value of `pointer-events`. That way you don't have to move it around

client/src/components/NavLink/NavLink.js
```diff
@@ -0,0 +1,43 @@
...
+      dropdownState: 'dropdown-hide'
+    }
+    this.routeTo = this.routeTo.bind(this);
+    this.showDropdown = this.showDropdown.bind(this);
+    this.hideDropdown = this.hideDropdown.bind(this);
+  }
+  // NEED to fix once routes are made
+  routeTo(url) {
+    return <Redirect to={url} />;
+  }
```
ZakkMan: You should be able to just use the `Link` component from React Router

client/src/components/NavLink/NavLink.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import "./NavLink.css";
+import NavLinkDropdownMenu from '../NavLinkDropdownMenu/NavLinkDropdownMenu';
+import { Redirect } from 'react-router-dom';
+
+class NavLink extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      dropdownState: 'dropdown-hide'
```
ZakkMan: This should be a boolean flag

client/src/components/NavLink/NavLink.js
```diff
@@ -0,0 +1,43 @@
...
+    }
+    this.routeTo = this.routeTo.bind(this);
+    this.showDropdown = this.showDropdown.bind(this);
+    this.hideDropdown = this.hideDropdown.bind(this);
```
ZakkMan: Can you create a single method (i.e. `toggleDropdown`)?

client/src/components/NavLink/NavLink.js
```diff
@@ -0,0 +1,43 @@
...
+  showDropdown() {
+    this.setState({
+      dropdownState: 'dropdown-show'
+    });
+  }
+  hideDropdown() {
+    this.setState({
+      dropdownState: 'dropdown-hide'
+    });
+  }
```
ZakkMan: Combine these into a single method

client/src/components/NavLink/NavLink.js
```diff
@@ -0,0 +1,43 @@
...
+  render() {
+    let classList = `NavLink`;
```
ZakkMan: If you only have the one class then just assign it to the `className` attribute (i.e. get rid of the extra variable)

client/src/components/NavLink/NavLink.js
```diff
@@ -0,0 +1,43 @@
...
+    return (
+      <div className={classList} onClick={this.routeTo(this.props.url)} onMouseEnter={this.showDropdown} onMouseLeave={this.hideDropdown}>
```
ZakkMan: Use a `Link` component

client/src/components/NavLink/NavLink.js
```diff
@@ -0,0 +1,43 @@
...
+        {this.props.text}
+        <div className={this.state.dropdownState}>
+          <NavLinkDropdownMenu menuList={this.props.menuList} />
+        </div>
```
ZakkMan: Note that not every menu item is going to have a dropdown menu

client/src/components/NavLink/NavLink.stories.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import NavLink from './NavLink';
+
+const placeholderTextProp = "COMMUNITY";
+const placeholderUrl = "community page route";
```
ZakkMan: Use some data that is like a URL so that we can test this out

client/src/components/NavLink/NavLink.stories.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import NavLink from './NavLink';
+
+const placeholderTextProp = "COMMUNITY";
```
ZakkMan: Capitalize text with CSS, otherwise screen readers will read this as an acronym

client/src/components/NavLink/NavLink.stories.js
```diff
@@ -0,0 +1,22 @@
...
+const placeholderDropdownProp = [
+  {
+    text: 'Explore People',
+    url: 'community people route'
+  },
+  {
+    text: 'Join Groups',
+    url: 'community groups route'
+  }
+]
```
ZakkMan: This test data is great - can you change the `url` properties to be URLs though?

client/src/components/NavLink/NavLink.stories.js
```diff
@@ -0,0 +1,22 @@
...
+
+storiesOf('NavLink', module)
+  .add("NavLink", () => (
+    <NavLink text={placeholderTextProp} url={placeholderUrl} menuList={placeholderDropdownProp} />
```
ZakkMan: You may have to rethink how you're passing in the dropdown menu - not every `NavLink` will have one.

client/src/components/NavLink/NavLink.stories.js
```diff
@@ -0,0 +1,22 @@
...
+    text: 'Explore People',
+    url: 'community people route'
+  },
+  {
+    text: 'Join Groups',
+    url: 'community groups route'
+  }
+]
+
+storiesOf('NavLink', module)
```
ZakkMan: Can you create a story for a `NavLink` that does not have a dropdown menu?

client/src/components/NavLinkDropdownMenu/NavLinkDropdownMenu.css
```diff
@@ -0,0 +1,14 @@
+.NavLinkDropdownMenu {
+  border: 2px solid gray;
```
ZakkMan: Can you use the CSS variable for this gray color?

client/src/components/NavLinkDropdownMenu/NavLinkDropdownMenu.css
```diff
@@ -0,0 +1,14 @@
...
+  margin: 0;
+  padding: 5px;
+  list-style: none;
+}
+.NavLinkDropdownMenu-li {
```
ZakkMan: You never want to put a tag name inside a css class or id - just use the tag name! So instead of something like this, just use `.NavLinkDropdownMenu li`

client/src/components/NavLinkDropdownMenu/NavLinkDropdownMenu.js
```diff
@@ -0,0 +1,30 @@
+import React, {Component} from 'react';
+import "./NavLinkDropdownMenu.css";
+import { Redirect } from 'react-router-dom';
+
+class NavLinkDropdownMenu extends Component{
+  constructor(props){
+    super(props);
+    this.state = {
+      classList: "NavLinkDropdownMenu"
+    };
```
ZakkMan: Can you move this class into the `className` attribute in your JSX and convert this to a function component? It wont have any state

client/src/components/NavLinkDropdownMenu/NavLinkDropdownMenu.js
```diff
@@ -0,0 +1,30 @@
...
+  routeTo(url) {
+    return <Redirect to={url} />;
+  }
+  render(){
+    return(
+      <ul className={this.state.classList}>
+        {this.props.menuList.map(li => 
+          <li className={this.state.classList + '-li'} onClick={this.routeTo(li.url)}>
+            {li.text}
+          </li>
```
ZakkMan: Can you use the React Router `Link` component here?

client/src/components/NavLinkDropdownMenu/NavLinkDropdownMenu.stories.js
```diff
@@ -0,0 +1,20 @@
...
+const placeholderDropdownProp = [
+  {
+    text: 'Explore People',
+    url: 'community people route'
+  },
+  {
+    text: 'Join Groups',
+    url: 'community groups route'
+  }
+]
```
ZakkMan: Same here re the sample data - please use test URLs

client/src/components/NavLinkDropdownMenu/NavLinkDropdownMenu.css
bhearn7: there's actually no gray border in the figma, but there is a box-shadow, so i'll remove the border and add the shadow

client/src/components/NavLink/NavLink.css
```diff
@@ -0,0 +1,34 @@
+.NavLink-Link {
```
ZakkMan: Can you update your classnames to use the BEM naming convention? There's documentation in the readme for it now

client/src/components/NavLink/NavLink.js
```diff
@@ -0,0 +1,42 @@
...
+        dropdownVisible: false
+      });
+    } else {
+      if (this.props.menuList.length) {
+        this.setState({
+          dropdownVisible: true
+        });
+      }
+    }
+  }
```
ZakkMan: Can you find a way to shorten this?

ZakkMan: Maybe use something like `!this.state.dropdownVisible`

client/src/components/NavLink/NavLink.js
```diff
@@ -0,0 +1,42 @@
...
+  // NEED to fix Link to= once routes are made
+  render() {
+    return (
+      <Link className='NavLink-Link' to={this.props.url}>
```
ZakkMan: Can you use `to` for the `url`?

client/src/components/NavLink/NavLink.js
ZakkMan: This does look waaaay better than before! Really great work!

client/src/components/NavLinkDropdownMenu/NavLinkDropdownMenu.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+import "./NavLinkDropdownMenu.css";
+
+const NavLinkDropdownMenu = (props) => (
+  <ul className='NavLinkDropdownMenu'>
+    {props.menuList.map(li => 
+      <Link to={li.url}>
```
ZakkMan: Use `to` here as well instead of `url`

client/src/components/NavLink/NavLink.css
```diff
@@ -0,0 +1,34 @@
...
+  pointer-events: none;
+}
+.NavLink__dropdown-show {
+  position: absolute;
+  top: 100%;
+  left: 24px;
+  width: calc(100% - 24px);
+  opacity: 1;
+  pointer-events: auto;
+}
```
ZakkMan: Is there a way you can move/remove the duplicative CSS here?

client/src/components/NavLink/NavLink.js
```diff
@@ -0,0 +1,36 @@
...
+    this.toggleDropdown = this.toggleDropdown.bind(this);
+  }
+  toggleDropdown() {
+    if (this.props.menuList.length) {
+      this.setState({
+        dropdownVisible: !this.state.dropdownVisible
+      });
+    }
+  }
+  // NEED to fix Link to= once routes are made
```
ZakkMan: Do we still need this comment?

Issue #146, opened 2019-12-09
Component: Card Issue


shelbyvjacobs:
`<Card />` is the parent component of the following card components. These are grouped together because of their similar format/styling/content. Use {props.children} to pass these into `<Card />`. 

The `Card` component will have three props:
* `path`: a file path or url path to a background
* `color`: a solid color for the background
* `pathOrColor`: a boolean that when `true` will use the `path` for the background and when `false` uses the color

## Children

* `<BookTrip />` - #141 
![Screen Shot 2019-12-09 at 12 17 58 PM](https://user-images.githubusercontent.com/53311285/70465325-b4d8b280-1a8e-11ea-803b-5de44109f7fe.png)
* `<TrendingHashtag />` - #60 
![Screen Shot 2019-12-09 at 2 19 42 PM](https://user-images.githubusercontent.com/53311285/70465440-f8cbb780-1a8e-11ea-842b-9ff09d6ba3ac.png)
* `<UpcomingTrips />` - Including TripCard, a trip details, and a link to the trip page
    <img width="300" alt="Screen Shot 2019-12-09 at 2 43 54 PM" src="https://user-images.githubusercontent.com/55462909/70467531-67127900-1a93-11ea-9cdf-270687445e35.png">

* `<UpcomingEvents />` - #140 
![Screen Shot 2019-12-09 at 12 15 39 PM](https://user-images.githubusercontent.com/53311285/70465495-16008600-1a8f-11ea-8064-fe5b36a33173.png)
* `<PreviousTrips />` - #60 
![Screen Shot 2019-12-09 at 2 19 36 PM](https://user-images.githubusercontent.com/53311285/70465437-f701f400-1a8e-11ea-96dc-789961125312.png)
* `<GroupCard />` - Including info on Group Name, Members, and Location. #39
* `<TripCard />` - Including info on Location, Month, & Year. #39 
![Screen Shot 2019-12-09 at 2 14 19 PM](https://user-images.githubusercontent.com/53311285/70465101-3714a700-1a8e-11ea-951e-bb7a308d4696.png)

* Cards that have no text but rather just a background image are the default `<Card />`.

If you have worked on these children and the names are different, that's okay! Use your name (as long as it's semantic, of course 😁 )

PR #157, opened 2019-12-09
Component: InfoCard Base PR

TylerEikenberg:
This component is going to be the base component for many other card components. #146 
Changes size and color based on props.


client/src/components/InfoCard/InfoCard.css
```diff
@@ -0,0 +1,39 @@
...
+  width: 100%;
+  min-width: 225px;
+  height: 155px;
+}
+
+.small {
+  width: 100%;
+  min-width: 225px;
+  height: 125px;
+}
```
ZakkMan: Can you namespace these? i.e. `info-card-small`

client/src/components/InfoCard/InfoCard.js
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import "./InfoCard.css";
+
+// Function based React Component
+const InfoCard = props => {
+  // Default Class to apply to Component
+  let className = `info-card-container ${props.variant} ${props.size}`;
```
ZakkMan: Can you do this in the `className` attribute and not create an extra variable?

client/src/components/InfoCard/InfoCard.stories.js
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import InfoCard from "./InfoCard";
+import image from "../../../../assets/images/card_medium1.png"; //here for example
+
+storiesOf("InfoCard", module)
+  .add("Info Card, Purple", () => <InfoCard variant="purple" size="large" />)
```
ZakkMan: `color` might be a better prop name than `variant`

client/src/components/InfoCard/InfoCard.js
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import "./InfoCard.css";
+
+// Function based React Component
+const InfoCard = props => {
```
ZakkMan: Should this just be called `Card`?

TylerEikenberg:
@ZakkMan added all of your suggestions, let me know what you think!

client/src/components/InfoCard/InfoCard.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import "./InfoCard.css";
+
+const Card = props => {
+  let { size = "large", color, picture } = props;
+  return (
+    <div className={`info-card-container info-card-${color} info-card-${size}`}>
+      {picture ? <img src={picture} alt="whoops"></img> : null}
```
ZakkMan: Could we have the `alt` text passed in as a prop?

TylerEikenberg: Changed file names and component name to card

TylerEikenberg:
@ZakkMan added the changes you suggested

TylerEikenberg:
@ZakkMan fixed classnames

PR #158, opened 2019-12-09
Forum comment

DallasLind:
Utilized CSS variables, but still need a helping hand when it comes to figuring out the pics are props. Will try to convene with Jenna or whoever is working images.


DallasLind:
I seem to be having an issue with my branches, I'll see if I can do something about that real quick

DallasLind:
Moving to a different branch, please disregard

PR #160, opened 2019-12-09
Post-Body PR update css names

torykling:
Updated css names to follow new conventions


client/src/components/PostBody/PostBody.css
```diff
@@ -5,46 +5,46 @@
   color: var(--black);
 }
 
-.post-body-top {
+.PostBody--top {
```
ZakkMan: Shouldn't these be `PostBody__top`?

torykling: fixed this in the other branch and just merged those changes into this branch

PR #161, opened 2019-12-09
PR Css name fix

torykling:
updated these. Does this look right?

### DIFF
--- a/client/src/components/CommentBody/CommentBody.css
+++ b/client/src/components/CommentBody/CommentBody.css
@@ -14,6 +14,6 @@
   white-space: nowrap;
 }
 
-.CommentBody__p span {
+.CommentBody p span {
   display: block;
 }
--- a/client/src/components/CommentBody/CommentBody.css
+++ b/client/src/components/CommentBody/CommentBody.css
@@ -1,19 +1,19 @@
 .CommentBody {
-    display: flex;
-    flex-direction: row;
-    justify-content: space-between;
-    align-items: flex-start;
-    width: 60%;
-    height: auto;
-    background-color: var(--gray-dark);
-    font-size: 14px;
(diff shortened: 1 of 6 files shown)


client/src/components/CommentBody/CommentBody.css
```diff
@@ -1,19 +1,19 @@
...
-    margin-left: 10px;
-    white-space: nowrap;
+.CommentBody__date {
+  margin-left: 10px;
+  white-space: nowrap;
 }
 
-p span {
-    display: block;
+.CommentBody__p span {
```
ZakkMan: Since `p` is a tag, we should just have `.CommentBody p span` as the selector

client/src/components/CommentBody/CommentBody.js
```diff
@@ -1,29 +1,27 @@
...
-      </p>
-      <div className="Date">
-        <p> 
+  return (
+    <div className="CommentBody">
+      <p>{body}</p>
+      <div className="CommentBody__date">
+        <p>
           <span>{date}</span>
+          <br></br>
```
ZakkMan: Please don't use the `<br>` tag! use CSS for this

client/src/components/CommentBody/CommentBody.js
```diff
@@ -1,29 +1,27 @@
...
 // Function based React Component
-const CommentBody = (props) => {
+const CommentBody = props => {
   // Default Class to apply to Component
-  // No props right now but when there is data we can use 
+  // No props right now but when there is data we can use
   // this.props.comment and this.props.date
   const body = props.body;
   const date = props.date;
   const time = props.time;
```
ZakkMan: Can you clean this up with destructuring?

PR #162, opened 2019-12-10
Component Banner PR

llucas314:
Banner takes in a prop of banner type and props.children.


client/src/components/Banner/Banner.css
```diff
@@ -0,0 +1,38 @@
+.banner{
```
ZakkMan: This should be `.Banner`

client/src/components/Banner/Banner.css
```diff
@@ -0,0 +1,38 @@
...
+}
+
+.Banner__header3{
+  margin: 0 auto;
+}
+
+input{
+  margin: 0 auto;
+}
+
```
ZakkMan: Can you namespace this, per the BEM naming convention?

client/src/components/Banner/Banner.css
```diff
@@ -0,0 +1,38 @@
...
+  text-align: left;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+}
+
+.Banner__header3{
+  margin: 0 auto;
+}
```
ZakkMan: Can you find semantic names for these? We never want to put numbers in our css classes

client/src/components/Banner/Banner.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import "./Banner.css";
+
+const Banner = props => {
+  let classlist = "banner";
+  return <div className={classlist} style={{
+    backgroundImage: `${props.bannerType}`,
+    height: `${props.bannerHeight}`, fontSize: `${props.bannerFontSize}`, lineHeight: `${props.bannerLineHeight}`
```
ZakkMan: This is cool, but font size and line height should be determined by the CSS. Could we do an enum style prop for height?

client/src/components/Banner/Banner.stories.js
```diff
@@ -0,0 +1,163 @@
...
+
+const handleClick = e => {
+  console.log("Clicked");
+};
+
+storiesOf("Banner", module)
+  .add("Home Banner", () => <Banner
+    bannerType='var(--home-banner)'
+    bannerHeight='380px'
+  />)
```
ZakkMan: Can you remove `banner` from these prop names? It's redundant, because they're all props of the `Banner` component

client/src/components/Banner/Banner.stories.js
```diff
@@ -0,0 +1,163 @@
...
+  .add("Home Banner with Welcome text", () =>
+    <Banner bannerType='var(--home-banner)' bannerHeight='380px' bannerText='Welcome,' bannerFontSize='80px' bannerLineHeight='120px'>
+      <div className="Banner__header">
+        <h2>Welcome, Justine</h2>
+      </div>
+    </Banner>
+  )
```
ZakkMan: The `Banner__header` part should be part of the `Banner` component, so that we can render this like this:

```jsx
<Banner>
  <h2>Welcome, Justine!</h2>
</Banner>
```

client/src/components/Banner/Banner.css
```diff
@@ -0,0 +1,28 @@
+.Banner{
+  padding: 0;
+  margin: 0 auto;
+  width: 100%;
+  width: 100vw;
+  height: 231px;
```
ZakkMan: You probably want the height to be determined by the content inside the banner, rather than have a set height

client/src/components/Banner/Banner.css
```diff
@@ -0,0 +1,28 @@
...
+  background-position: center;
+  background-size: cover;
+  background-repeat: no-repeat;
+  text-align: center;
+  font-family: var(--default-font);  
+  font-size: 48px;
+  font-style: normal;
+  font-weight: normal;
+  color: #FAFAFA;
```
ZakkMan: Can you use the css variables for color and font properties?

client/src/components/Banner/Banner.css
```diff
@@ -0,0 +1,28 @@
...
+  line-height: 72px;
+}
+
+.Banner__header3{
+  margin: 10px auto;
+}
```
ZakkMan: Why does this have a `3` in the name?

client/src/components/Banner/Banner.css
```diff
@@ -0,0 +1,28 @@
...
+
+.Banner .InputTextField__container--search{
+  margin: 0 auto;
+}
```
ZakkMan: We never want to affect the CSS inside another component. If you need to change the look of another component, then you need to update that component

client/src/components/Banner/Banner.js
```diff
@@ -0,0 +1,16 @@
...
+import "./Banner.css";
+
+const Banner = props => {
+  let classlist = "Banner";
+  return <div className={classlist} style={{
+    backgroundImage: `url(${props.background})`,
+    height: `${props.Height}`, fontSize: `${props.FontSize}`, lineHeight: `${props.LineHeight}`,
+    textAlign: `${props.TextAlign}`, display: `${props.display}`, flexDirection: `${props.flexdirection}`,
+    justifyContent: `${props.justifycontent}`, alignItems: `${props.alignitems}`
+  }}>
```
ZakkMan: Do you need to pass all these props down? This seems excessive. Remember that you can apply styling to the children directly!

ZakkMan: We definitely want to apply the background image this way

client/src/components/Banner/Banner.stories.js
```diff
@@ -0,0 +1,129 @@
...
+    <Banner background={Banner__purpleBig} Height='380px' FontSize='80px' LineHeight='120px' TextAlign="left" display='flex' flexdirection='column'
+    justifycontent='center' alignitems='center'> 
+      <h2>Welcome Back,<br />Justine.</h2>
+    </Banner>
+  )
+
+  .add("Community Banner", () => <Banner background={Banner__purpleBig}/>)
+  .add("Community Banner with input People", () =>
+    <Banner background={Banner__purpleBig}>
+        <h3 className="Banner__header3"> Community: People </h3>
```
ZakkMan: We're not going to remember this class is here, so it'd be better to remove it from the CSS and apply these styles directly to this `h3` tag

PR #168, opened 2019-12-10
PR  [updated README with BEM]

tissakhosla:


client/README.md
```diff
@@ -62,3 +62,32 @@ To generate a **class** component, run the following
 ```sh
 hygen component-class new --name NameOfComponent
 ```
+### CSS Class Names
+
+- To avoid style conflicts we want to make sure all our CSS selectors are unique.
+- This will be the naming convention, based on BEM. [https://css-tricks.com/bem-101/]
```
ZakkMan: Can you turn this into a Markdown link?

client/README.md
```diff
@@ -62,3 +62,32 @@ To generate a **class** component, run the following
...
+```
+.ComponentName
+.ComponentName--modifier
+.ComponentName__elementName
+.ComponentName__elementName--modifier
```
ZakkMan: Did y'all decide on `elementName` over `element-name`? I think the PR I merged is using `element-name`

wchrapcynski: Just a note. I personally like element-name and keeping the camel case for JS just to have things separate.

client/README.md
ZakkMan: @tissakhosla did you miss this one?

Issue #155, opened 2019-12-09
Component: Burger icon


wchrapcynski:
A burger icon is needed for the mobile version. This is not part of the mock but it will be needed. When not clicked(engaged) it will be three lines made from divs. When clicked, the middle div is hidden and the other two form an X.

Props:
*Clicked
*Callback function for menu

wchrapcynski:
This is more or less done. It will work in conjunction with the navbar which I'm also working on. Will submit for a pull request once it's all working.

wchrapcynski:
This is complete.

PR #169, opened 2019-12-10
Burger component PR

wchrapcynski:
Burger should be ready to go per issue #155. May need to adjust size once it's in the Navbar. Call back function is set up but in a comment.


client/src/components/Burger/Burger.css
```diff
@@ -0,0 +1,42 @@
...
+    height: 8px;
+    width: 50px;
+    background-color: var(--black);
+    margin: 8px 0;
+    transition: .2s ease-in-out;
+    -webkit-transition: .1s ease-in-out;
+    -moz-transition: .2s ease-in-out;
+    -ms-transition: .2s ease-in-out;
+    -o-transition: .2s ease-in-out;
+}
```
ZakkMan: Can we give these all a shared class? Something like `.burger-line` would be good.

client/src/components/Burger/Burger.css
```diff
@@ -0,0 +1,42 @@
...
+    -webkit-transform: translate(0, -16px) rotate(-45deg);
+    -moz-transform: translate(0, -16px) rotate(-45deg);
+    -ms-transform: translate(0, -16px) rotate(-45deg);
+    -o-transform: translate(0, -16px) rotate(-45deg);
+    transition:  .1s ease-in-out;
+    -webkit-transition:  .1s ease-in-out;
+    -moz-transition:  .1s ease-in-out;
+    -ms-transition:  .1s ease-in-out;
+    -o-transition:  .1s ease-in-out;
+}
```
ZakkMan: Rather than name these `burger-line-one` and `burger-line-three` can you use a pseudoclass? (i.e. `.burger-line:first-of-type`)

client/src/components/Burger/Burger.css
```diff
@@ -0,0 +1,42 @@
+.burger-line-one, .burger-line-two, .burger-line-three {
```
ZakkMan: I think you need a parent class, like `.burger`

Typically, you'd have something like `.burger` and `.burger--active` to distinguish between the different states.

ZakkMan: Then you can modify the styling for the burger lines based on whether or not that `.burger--active` class is there

client/src/components/Burger/Burger.js
```diff
@@ -0,0 +1,55 @@
+import React, { Component } from "react";
+import "./Burger.css";
+
+// Class Based React Component
+class Burger extends Component {
+  constructor(props) {
+    super(props);
+    console.log(props);
```
ZakkMan: Can you remove any `console.log` statements before making a PR from now on?

client/src/components/Burger/Burger.js
```diff
@@ -0,0 +1,55 @@
...
+
+    // Default CSS class to apply to the Component
+    this.state = {
+      classList: "Burger",
```
ZakkMan: Can you move this class into the `className` attribute?

client/src/components/Burger/Burger.js
```diff
@@ -0,0 +1,55 @@
...
+  };
+
+  // Runs after Component is loaded in the broswer
+  componentDidMount() {}
+
+  // Runs after a component has been updated
+  componentDidUpdate() {}
+
+  // Runs right before a component is removed from the DOM
+  componentWillUnmount() {}
```
ZakkMan: Can you remove these if you're not using them?

client/src/components/Burger/Burger.js
```diff
@@ -0,0 +1,55 @@
...
+    this.state = {
+      classList: "Burger",
+      active: false
+    };
+  }
+
+  clicked = () => {
+    this.setState({ active: !this.state.active })
+    // Utilizes a callback in the Navbar to display the menu.
+    // this.props.DisplayMenu()
```
ZakkMan: Can you set this up to take and invoke the a callback?

client/src/components/Burger/Burger.js
```diff
@@ -0,0 +1,55 @@
...
+  componentDidUpdate() {}
+
+  // Runs right before a component is removed from the DOM
+  componentWillUnmount() {}
+
+  render() {
+    console.log(this.state.active);
+    return (
+      <div>
+        <div className={this.state.classList} onClick={this.clicked}>
```
ZakkMan: Can we give the method a more semantic name? i.e. `onClick`?

client/src/components/Burger/Burger.stories.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import Burger from './Burger';
+
+storiesOf('Burger', module)
+  .add("Default", () => (
+    <Burger clicked=""/>
```
ZakkMan: Can you pass in a callback here?

client/src/components/Burger/Burger.js
```diff
@@ -0,0 +1,55 @@
+import React, { Component } from "react";
+import "./Burger.css";
+
+// Class Based React Component
+class Burger extends Component {
```
ZakkMan: I'm not sure this needs to be a class component - I think you could accept `active` as a prop to determine whether or not it's active or not. The callback passed in should toggle active/inactive as a boolean (true for active, false for inactive)

client/src/components/Navbar/Navbar.css
```diff
@@ -0,0 +1,24 @@
+
+.Navbar {
```
ZakkMan: Should this navbar stuff be in a different branch?

wchrapcynski: Hrm. It should be. I might have used git add -A by accident.

client/src/components/Burger/Burger.css
wchrapcynski: I had them separate because something slightly different happens to each one. When one and three are animated, the rotating is different. Two is the only one that disappears.

wchrapcynski: Nevermind. I see these can be all one now. I'll fix it.

wchrapcynski: Done.

client/src/components/Burger/Burger.css
wchrapcynski: This has been updated per the new naming convention.

client/src/components/Burger/Burger.js
wchrapcynski: Removed.

client/src/components/Burger/Burger.js
wchrapcynski: This should now be set up as we discussed.

client/src/components/Burger/Burger.css
```diff
@@ -0,0 +1,41 @@
...
+    height: 8px;
+    width: 50px;
+    background-color: var(--black);
+    margin: 8px 0;
+    transition: .1s ease-in-out;
+    -webkit-transition: .1s ease-in-out;
+    -moz-transition: .1s ease-in-out;
+    -ms-transition: .1s ease-in-out;
+    -o-transition: .1s ease-in-out;
+}
```
ZakkMan: Can you move this to the `.Burger__line` class? That will be more explicit

wchrapcynski: Done.

PR #173, opened 2019-12-10
Component: Input text field, styling changes PR

TylerEikenberg:
Hey @ZakkMan I had to add some styling changes for this component, it was having centering issues. I also added the more explicit class names.


client/src/components/InputTextField/InputTextField.css
```diff
@@ -1,17 +1,16 @@
-.InputTextField--default input {
+.InputTextField__container-default {
```
ZakkMan: It looks like `default` is a modifier here. If so, it should be `InputTextField__container--default`

TylerEikenberg: Fixed per your comment

PR #174, opened 2019-12-10
Button CSS w/ block-element-modifier PR

llucas314:
I update CSS with block-element-modifier.

### DIFF
--- a/client/src/components/Button/Button.css
+++ b/client/src/components/Button/Button.css
@@ -15,47 +15,47 @@
 .button:hover {
   background-color: #fcf0ee;
 }
-.button-large {
+.button--large {
   padding: 0.13em 2.71em;
   font-size: 36px;
 }
-.button-small {
+.button--small {
   padding: 0.42em 3.35em;
   font-size: 18px;
 }
-.button-pink {
+.button__pink {
   color: var(--white);
   background-color: var(--pink);
 }
-.button-pink:hover {
+.button__pink:hover {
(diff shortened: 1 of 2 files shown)


client/src/components/Button/Button.css
```diff
@@ -15,47 +15,47 @@
 .button:hover {
   background-color: #fcf0ee;
 }
-.button-large {
+.button--large {
```
torykling: This looks great! Can we change all the bs to be uppercase for consistency?

.Button--large

client/src/components/Button/Button.css
```diff
@@ -15,47 +15,47 @@
...
   padding: 0.13em 2.71em;
   font-size: 36px;
 }
-.button-small {
+.button--small {
   padding: 0.42em 3.35em;
   font-size: 18px;
 }
-.button-pink {
+.button__pink {
```
torykling: This was my fault. It was not clear in the presentation. 

A pink variant of the component would be .Button--pink
I think you only have variants here so all of these should be -- not __
In my example, I had a subsection of my component that was purple. So I had .Component__purple. That was confusing. sorry!

client/src/components/Button/Button.css
```diff
@@ -15,47 +15,47 @@
 .button:hover {
```
torykling: This looks great! Can we change all the bs to uppercase for consistency?

.Button:hover

PR #177, opened 2019-12-10
Component: Trending Hashtag Card

TylerEikenberg:
This component uses the component ```Card``` as a parent. Pass in ```data``` prop which will be an array of hashtags.


client/src/components/TrendingHashtagCard/TrendingHashtagCard.css
```diff
@@ -0,0 +1,40 @@
+.TrendingHashtagCard_container {
```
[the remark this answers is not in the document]
TylerEikenberg: fixed!

TylerEikenberg:
@ZakkMan Added your suggestions!

client/src/components/TrendingHashtagCard/TrendingHashtagCard.css
```diff
@@ -0,0 +1,40 @@
...
+  flex-direction: column;
+  flex-wrap: wrap;
+  align-items: center;
+  position: absolute;
+  left: 50%;
+  top: 65%;
+  width: 350px;
+  height: 200px;
+  transform: translate(-50%, -50%);
+}
```
ZakkMan: Just noticed the absolute positioning here - can you find a way to do this with flexbox? We typically want to avoid absolute positioning, except for special cases

TylerEikenberg: @ZakkMan removed all the absoulte positioning, let me know what you think!

TylerEikenberg:
@ZakkMan fixed the styling issue with small amounts of data passed in and added a conditional to trim down the data if its more than 10. 
Let me know what you think!

Issue #140, opened 2019-12-09
Component: UpcomingEvents Issue


shelbyvjacobs:
The `UpcomingEvents` component is a child of the `Card` component #146. It should have four props:
* `name`: the name of the event
* `date`: a string formatted "MONTH 00th, YEAR", year is optional
* `time`: a string with the time formatted "0:00 PM"
* `location`: the city or area where the event is occurring 

### Examples
![Screen Shot 2019-12-09 at 12 15 39 PM](https://user-images.githubusercontent.com/53311285/70457179-a9c95680-1a7d-11ea-9fb2-a00e3bd6037f.png)

PR #180, opened 2019-12-10
Component: Upcoming Events Card PR

shelbyvjacobs:
From Issue #140


client/src/components/UpcomingEventsCard/UpcomingEventsCard.css
```diff
@@ -0,0 +1,28 @@
...
+.UpcomingEventsCard__date,
+.UpcomingEventsCard__time,
+.UpcomingEventsCard__location {
+    font-family: var(--paragraph-font);
+    font-size: 1.2vw;
+}
+
+.UpcomingEventsCard__location {
+    padding-top: .8vh;
+}
```
erinkelley27: You may want to add some padding to the bottom. The location is going over the bottom edge of the card.

client/src/components/UpcomingEventsCard/UpcomingEventsCard.css
```diff
@@ -0,0 +1,28 @@
+.UpcomingEventsCard {
+    width: 17vw;
+    height: 14vh;
+    border-radius: 3px;
+    background-color: var(--pink);
+    filter: drop-shadow(0px 4px 4px rgba(0, 0, 0, 0.25));
```
erinkelley27: Is there a CSS variable for this color? It looks a little lighter than the `Info Card, Pink, Small` card's shadow.

client/src/components/UpcomingEventsCard/UpcomingEventsCard.js
```diff
@@ -0,0 +1,18 @@
...
+
+const UpcomingEventsCard = (props) => {
+
+  return(
+    <Link className="UpcomingEventsCard" to="#">
+      <span className="UpcomingEventsCard__name">{props.name}</span>
+      <span className="UpcomingEventsCard__date">Date: {props.date}</span>
+      <span className="UpcomingEventsCard__time">Time: {props.time}</span>
+      <span className="UpcomingEventsCard__location">Location: {props.location}</span>
+    </Link>
```
erinkelley27: Try importing the `Info Card, Pink, Small` card here and wrapping the link around that. From there, you can add the name, date, time and location properties. The `Info Card, Pink, Small` should have a min-width and other CSS properties that your `UpcomningEventsCard` can inherit so that your location isn't running off the bottom edge of the card.

shelbyvjacobs: Is there a way to import just the `Info Card, Pink, Small`? Or do I import all of the Card parent component and just add the `color` and `size` attributes in UpcomingEventsCard.js? I'm not quite sure how to go about this.

erinkelley27:
@shelbyvjacobs Great question - sorry that wasn't clear. Yes, you can import the Card component and then pass in the props for that small pink card. Does that make sense?

shelbyvjacobs:
Yes it does! It should be ready to review again, in that case. Thank you!

client/src/components/UpcomingEventsCard/UpcomingEventsCard.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import "./UpcomingEventsCard.css";
+// import { prependOnceListener } from 'cluster';
```
erinkelley27: Removed commented out code.

client/src/components/UpcomingEventsCard/UpcomingEventsCard.css
```diff
@@ -0,0 +1,27 @@
...
+    height: 100%;
+    border-radius: 3px;
+    background-color: var(--pink);
+    filter: drop-shadow(0px 4px 4px rgba(0, 0, 0, 0.25));
+    color: var(--white);
+    display: flex;
+    flex-direction: column;
+    padding: 10px;
+    text-decoration: none;
+}
```
erinkelley27: How much of this is the same as the Card component? Can you eliminate some of this CSS?

shelbyvjacobs:
Cool! I remembered the excess CSS right after I requested another review. 🤦‍♀ Sorry! That's sorted out and the commented code has been removed.

Issue #179, opened 2019-12-10
Component: Like Comment Buttons Issue


torykling:
![image](https://user-images.githubusercontent.com/55210397/70550613-04c58100-1b44-11ea-8362-017cf6e3d80a.png)

Location: forum post

PR #181, opened 2019-12-10
Component button: Like and Comment Buttons PR

torykling:
I added like and comment buttons for new issue #179. 

I set a min-width for the buttons but set their width to generally be 50% of the container because I want their width to expand to fill half of the post container. Does that seem right?

### DIFF
--- a/client/src/components/Button/Button.css
+++ b/client/src/components/Button/Button.css
@@ -11,6 +11,7 @@
   transition-duration: 0.3s;
   -webkit-transition-property: background-color;
   transition-property: background-color;
+  white-space: nowrap;
 }
 .Button:hover {
   background-color: #fcf0ee;
--- a/client/src/components/Button/Button.css
+++ b/client/src/components/Button/Button.css
@@ -14,7 +14,7 @@
   white-space: nowrap;
 }
 .Button:hover {
-  background-color: #fcf0ee;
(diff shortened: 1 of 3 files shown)


client/src/components/Button/Button.stories.js
```diff
@@ -73,4 +75,12 @@ storiesOf("Button", module)
       textColor={"black"}
       handleClick={handleClick}
     />
+  ))
+  .add("Like Button", () => <Button icon={heart} text="Like" type="like" />)
+  .add("Comment Button", () => <Button icon={comment} text="Comment" type="comment" />)
+  .add("contained", () => (
```
erinkelley27: Change the name of this story to be more descriptive of what it actually is.

client/src/components/Button/Button.css
```diff
@@ -11,6 +11,7 @@
   transition-duration: 0.3s;
   -webkit-transition-property: background-color;
   transition-property: background-color;
+  white-space: nowrap;
 }
 .Button:hover {
   background-color: #fcf0ee;
```
erinkelley27: Is there a CSS variable for this color? If not, should there be?

torykling:
thanks Erin! changed the title and var names for hover colors. Bill is adding them to the app css now.

PR #201, opened 2019-12-11
Component: Card,  Update PR

TylerEikenberg:
Screen Shot 2019-12-10 at 7 04 32 PM

I realized we were missing this version of the card component from the mock-up so I put it together really quick. Let me know if you guys see anything that could be changed. To render this version of the card you pass in a prop of ```discover={true}```

### DIFF
--- a/client/src/components/Card/Card.js
+++ b/client/src/components/Card/Card.js
@@ -2,10 +2,10 @@ import React from "react";
 import "./Card.css";
 
 const Card = props => {
-  let { size = "large", color, picture, alt } = props;
+  let { size = "large", color, picture, discover } = props;
   return (
     <div
-      className={`Card Card--${color} Card--${size}`}
+      className={`Card Card--${color} Card--${size} Card__discover-${discover}`}
       style={{ backgroundImage: `url(${picture})` }}
     >
       {props.children}
--- a/client/src/components/Card/Card.js
+++ b/client/src/components/Card/Card.js
@@ -2,7 +2,7 @@ import React from "react";
 import "./Card.css";
 
 const Card = props => {
-  let { size = "large", color = "purple", picture = null, outline = false } = props;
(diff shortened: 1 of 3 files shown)


client/src/components/Card/Card.js
```diff
@@ -2,10 +2,10 @@ import React from "react";
 import "./Card.css";
 
 const Card = props => {
-  let { size = "large", color, picture, alt } = props;
+  let { size = "large", color, picture, discover } = props;
   return (
     <div
-      className={`Card Card--${color} Card--${size}`}
+      className={`Card Card--${color} Card--${size} Card__discover-${discover}`}
```
ZakkMan: Note that without a default value, this could break the CSS classname (we could end up with a class name of `Card__discover-` (same actually goes for the `size` value!

client/src/components/Card/Card.stories.js
```diff
@@ -4,13 +4,18 @@ import Card from "./Card";
...
+  .add("Card, Pink", () => <Card color="pink" size="large" />)
+  .add("Card, Pink, Small", () => <Card color="pink" size="small" />)
+  .add("Card, Blank", () => <Card size="large" />)
+  .add("Card, Blank, Small", () => <Card size="small" />)
+  .add("Card, Picture BG, Large", () => <Card picture={image} alt="whoops" size="medium" />)
+  .add("Card, Picture Hawaii, Large", () => (
     <Card picture={"https://wallpaperaccess.com/full/144067.jpg"} alt="whoops" size="medium" />
+  ))
+  .add("Card, Discover New Groups", () => (
+    <Card discover={true} size="medium">
```
ZakkMan: Could we call the prop something more semantic and less `contextual`? i.e. `border`, `outline` or something like that?

TylerEikenberg:
@ZakkMan changes added! 👍

PR #203, opened 2019-12-11
PR: Component extra users

tissakhosla:
This component is in the inbox, and could be used in My Connections/Her Connections/Members, which is the component 'Connections'


client/src/App.css
```diff
@@ -48,7 +48,7 @@
     --social-instagram: url("../../assets/images/social_instagram.png");
     --social-pinterest: url("../../assets/images/social_pinterest.png");
     --splash-background: url("../../assets/images/splash_background.png");
-    --text-box-search: url("../../assets/images/text_box_search.png");
+    /* --text-box-search: url("../../assets/images/text_box_search.png"); */
```
ZakkMan: can you delete commented out code?

client/src/components/ExtraUsers/ExtraUsers.js
```diff
@@ -0,0 +1,54 @@
...
+import React, { Component } from 'react';
+import { Link } from "react-router-dom";
+import "./ExtraUsers.css";
+
+
+// Class Based React Component
+class ExtraUsers extends Component {
+  constructor(props) {
+    super(props);
+    console.log(props);
```
ZakkMan: Can you remove any `console.log` statements?

client/src/components/ExtraUsers/ExtraUsers.js
```diff
@@ -0,0 +1,54 @@
+
+import React, { Component } from 'react';
+import { Link } from "react-router-dom";
+import "./ExtraUsers.css";
+
+
+// Class Based React Component
+class ExtraUsers extends Component {
```
ZakkMan: I don't know that this component needs state - I think it could be a function component

client/src/components/ExtraUsers/ExtraUsers.js
```diff
@@ -0,0 +1,54 @@
...
+  componentDidUpdate() { }
+
+
+  // Runs right before a component is removed from the DOM
+  componentWillUnmount() { }
+
+  render() {
+    return (
+      <div className='ExtraUsers'>
+        <Link style={{textDecoration: 'none'}} to={this.props.to} className='ExtraUsers__text'>
```
ZakkMan: Can you move `textDecoration` to the CSS

client/src/components/ExtraUsers/ExtraUsers.js
```diff
@@ -0,0 +1,33 @@
...
+// Class Based React Component
+function ExtraUsers(props) {
+  let extraUsersValue = null
+
+  function howManyMore(theArray) {
+    if (theArray.length > 4) {
+      let extraValue = theArray.length - 4
+      return extraValue
+    }
+  }
```
ZakkMan: This doesn't need to be in a function. Shouldn't this only get used by the parent component if there are extras? In that case, this should just take the number of extra users and render that number - keep it simple!

client/src/components/ExtraUsers/ExtraUsers.css
```diff
@@ -0,0 +1,18 @@
+.ExtraUsers {
+  background: var(--purple);
+  height: 150px;
+  width: 150px;
```
erinkelley27: Check with the responsive team to see if these values need to be in rem instead of px.

Issue #216, opened 2019-12-11
Onboarding Parent Component 


MobolanleAdebesin:
In order to keep a consistent structure and formatting across all 3 onboarding pages. We created a Parent component called OnboardingParent That includes the logo, CarouselDots, and styling for every onboarding page.

PR #217, opened 2019-12-11
PR: Parent Onboarding Component

MobolanleAdebesin:
Setup parent Onboarding component to act as a template for Onboarding children components.
#216

### DIFF
--- a/client/src/components/OnboardingParent/OnboardingParent.js
+++ b/client/src/components/OnboardingParent/OnboardingParent.js
@@ -2,7 +2,7 @@ import React from "react";
 import "./OnboardingParent.css";
 import Logo from "../Logo/Logo.js";
 import CarouselDots from "../CarouselDots/CarouselDots";
-// Class Based React Component
+
 const OnboardingParent = props => {
   return (
     <div className="OnboardingParent">
--- a/client/src/components/OnboardingParent/OnboardingParent.js
+++ b/client/src/components/OnboardingParent/OnboardingParent.js
@@ -7,9 +7,9 @@ const OnboardingParent = props => {
   return (
     <div className="OnboardingParent">
       <div className="OnboardingParent__wrapper">
-        <Logo />
+        <Logo className="OnboardingParent__logo" />
         {props.children}
-        <div className="test">
+        <div className="OnboardingParent__carousel-dots-container">
           <CarouselDots
             numberOfDots={3}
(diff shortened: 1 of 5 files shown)


client/src/components/OnboardingParent/OnboardingParent.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  // Runs after Component is loaded in the broswer
+  componentDidMount() {}
+
+  // Runs after a component has been updated
+  componentDidUpdate() {}
+
+  // Runs right before a component is removed from the DOM
+  componentWillUnmount() {}
```
erinkelley27: You don't need lifecycle methods if you don't have functionality that you specifically want to attach to a stage in the component lifecycle. Think about it this way - if you're not using them, you don't need to include them.

client/src/components/OnboardingParent/OnboardingParent.js
```diff
@@ -0,0 +1,46 @@
...
+
+  render() {
+    return (
+      <div className={this.state.classList}>
```
erinkelley27: Since the state doesn't seem to be changing in this component, you could probably just use the actual className of `OnboardingParent` here directly. Then, you can make this a functional component.

client/src/components/OnboardingParent/OnboardingParent.js
```diff
@@ -0,0 +1,46 @@
...
+        <div className="OnboardingParent__wrapper">
+          <img src={logo} alt="co-tripper-logo" />
```
erinkelley27: Can you import a logo component here instead of a static image?

MobolanleAdebesin: I imported the logo from the Logo component, but I am unable to increase the size.

client/src/components/OnboardingParent/OnboardingParent.js
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import "./OnboardingParent.css";
+import Logo from "../Logo/Logo.js";
+import CarouselDots from "../CarouselDots/CarouselDots";
+// Class Based React Component
```
erinkelley27: This is no longer a class based component - will you update this comment?

stuvun: Done

Issue #197, opened 2019-12-10
Responsive Splash Page


joshseven:

bhearn7:
Kid Responsive reporting for duty

PR #249, opened 2019-12-11
responsive splash page

bhearn7:
per issue #197 

1. started the responsive splash page

2. commented out the navbar placeholder tory made so splash page would fit fullscreen

NOTE 1: per the figma: normal white Login button needed to be same size as normal purple Register button, so I added some custom css (.SplashPage__button-margin--small a button)

NOTE 2: Button component requires handleClick function as a prop so I just defined a dummy function that console logs some text since these splash page buttons will be wrapped in a Link

NEED: pages & routes to be made for Login & Register


client/src/App.js
```diff
@@ -1,21 +1,18 @@
...
   render() {
     return (
       <div className='App'>
-        <nav>
+        {/* <nav>
           <Link to="/">Splash</Link>
-        </nav>
+        </nav> */}
         <main>
           <Route path="/" exact render={this.splash}></Route>
```
ZakkMan: This isn't really how React Router's `render` work. And, since you're just rendering a single component, could you use the `component` property for now?

client/src/components/NavLinkDropdownMenu/NavLinkDropdownMenu.js
```diff
@@ -5,7 +5,7 @@ import "./NavLinkDropdownMenu.css";
 const NavLinkDropdownMenu = (props) => (
   <ul className='NavLinkDropdownMenu'>
     {props.menuList.map(li => 
-      <Link to={li.to}>
+      <Link to={li.to} key={li.text}>
         <li>{li.text}</li>
       </Link>
     )}
```
ZakkMan: This should probably be fixed on the branch for this component, rather than in this PR

ZakkMan: It's fine here - going forward though, you want to keep your PRs relatively atomic (only focused on the change you're requesting be merged)

client/src/components/SplashPage/SplashPage.js
```diff
@@ -3,13 +3,20 @@ import { Link } from 'react-router-dom';
...
       <h1>CoTripper</h1>
       <p>Book trips, meet moms, and experience family travel together!</p>
-      <Link to='/login'><Button text="Login" /></Link>
-      <Link to='/register'><Button text="Register" color="purple" /></Link>
+      <div className='SplashPage__button-margin--small'>
+        <Link to='/login'><Button text="Login" handleClick={handleClick}/></Link>
+      </div>
+      <div className='SplashPage__button-margin'>
+        <Link to='/register'><Button text="Register" color="purple" handleClick={handleClick}/></Link>
+      </div>
```
ZakkMan: Is it possible to use the `Button` component here? That would be better

client/src/index.css
```diff
@@ -1,13 +1,19 @@
...
   -moz-osx-font-smoothing: grayscale;
 }
-
 code {
   font-family: source-code-pro, Menlo, Monaco, Consolas, "Courier New",
     monospace;
 }
+#root {
+  height: 100%;
+}
```
ZakkMan: What are you trying to do here with all these `height: 100%`? Does the `body` or `root` element need to take up the whole screen? If so, then you should put `min-height: 100vh` on one of them (whichever one makes the most sense)

bhearn7:
@ZakkMan removed my bogus height: 100% stuff as well

bhearn7:
@ZakkMan on it!

bhearn7:
@ZakkMan probably should of added the pages folder in a 'hotfix-pages' branch but oh well

PR #266, opened 2019-12-11
User model & Profile model

tully4school:
Done and migrated


api/cotrip_app/models.py
```diff
@@ -1,3 +1,28 @@
 from django.db import models
 
+
+class User(models.Model):
```
erinkelley27: Django actually comes with a User model already. You can use that one instead of creating your own.

api/cotrip_app/models.py
```diff
@@ -1,3 +1,28 @@
...
+    email = models.EmailField(max_length=200)
+    password = models.CharField(max_length=100)
+
+    def __str__(self):
+        # return super().__str__()
+        return self.email
+
+
+class Profile(models.Model):
```
erinkelley27: Here is a really helpful article about extending the User model to a Profile model. Read through this - it may inspire you to make some adjustments on how you want to set this up. Instead of a one-to-many relationship, you will want the User and Profile to have a one-to-one relationship.

[How to Extend Django User Model](https://simpleisbetterthancomplex.com/tutorial/2016/07/22/how-to-extend-django-user-model)

api/cotrip_app/models.py
```diff
@@ -1,3 +1,28 @@
...
+    city_of_residence = models.CharField(max_length=200)
+    dream_destination = models.CharField(max_length=100)
+    profile_photo = models.ImageField(max_length=200)
+    # topics = models.ForeignKey(Topic, on_delete=models.PROTECT)
+
+    def __str__(self):
+        return self.last_name
+
+
 # Create your models here.
```
erinkelley27: Remove any commented out code that you don't need.

_templates/generator/help/index.ejs.t
```diff
@@ -0,0 +1,5 @@
+---
+message: |
+  hygen {bold generator new} --name [NAME] --action [ACTION]
+  hygen {bold generator with-prompt} --name [NAME] --action [ACTION]
+---
```
ZakkMan: @sarahtsige Can you reset these files in the `_templates` directory? Not sure why they're showing up in this pull request.

sarahtsige: @ZakkMan  by reset, do you mean delete these lines?

Issue #69, opened 2019-12-06
Component: Molecular Onboarding Step 1 Issue


shelbyvjacobs:
![Screen Shot 2019-12-06 at 2 12 51 PM](https://user-images.githubusercontent.com/53311285/70349349-924a5d80-1832-11ea-9dc7-79c42f901200.png)

PR #282, opened 2019-12-12
Onboarding1 component PR

MobolanleAdebesin:
#69


client/src/components/OnboardingOne/OnboardingOne.js
```diff
@@ -0,0 +1,57 @@
...
+
+    this.state = {
+      classList: "OnboardingOne"
+    };
+  }
+
+  render() {
+    return (
+      <OnboardingParent>
+        <div className={this.state.classList}>
```
erinkelley27: If you're not updating state in this component, does it need to be a class component? Do you need to set the class from state?

client/src/components/OnboardingParent/OnboardingParent.js
```diff
@@ -4,18 +4,16 @@ import Logo from "../Logo/Logo.js";
 import CarouselDots from "../CarouselDots/CarouselDots";
 
 const OnboardingParent = props => {
+  const hello = () => {
+    console.log("Hello world!");
+  };
```
erinkelley27: Change this function to something more semantic. For example, you could call it `handleStatusChange` and have it console log a string indicating that the status has changed.

bkeithbii:
> @bkeithbii Can you fix the possible merge conflicts before I review this? I think you'll just need to update this branch with the develop branch (i.e. merge develop into this branch)

I made some fixes, says I need to commit first to merge. There are 232 staged changes once I checked out to this branch, should I go ahead and commit all of those??

client/src/components/OnboardingOne/OnboardingOne.js
```diff
@@ -0,0 +1,55 @@
...
+
+class OnboardingOne extends Component {
+  constructor(props) {
+    super(props);
+    console.log(props);
+
+    this.state = {
+      classList: "OnboardingOne"
+    };
+  }
```
ZakkMan: Can this be a function component?

client/src/components/OnboardingOne/OnboardingOne.js
```diff
@@ -0,0 +1,55 @@
+import React, { Component } from "react";
+import "./OnboardingOne.css";
+import InputTextField from "../InputTextField/InputTextField.js";
+import Button from "../Button/Button.js";
+import OnboardingParent from "../OnboardingParent/OnboardingParent";
+
+class OnboardingOne extends Component {
```
ZakkMan: Can you think of a better name for this than `OnboardingOne`?

client/src/components/OnboardingOne/OnboardingOne.css
```diff
@@ -0,0 +1,58 @@
...
+  .OnboardingOne__header {
+    font-size: 1.5rem;
+  }
+  .OnboardingOne .InputTextField__container--default {
+    min-width: 11.5rem;
+    max-width: 11.5rem;
+    font-size: 1rem;
+    height: 0.5rem;
+    padding: 0.7rem 1.4rem;
+  }
```
ZakkMan: You shouldn't modify the styling of a separate component like this. This will make debugging really difficult later. Instead, if that component isn't doing something you need, then you need to update that component

client/src/components/OnboardingOne/OnboardingOne.css
```diff
@@ -0,0 +1,58 @@
...
+  width: 22.5rem;
+}
+
+.OnboardingOne .Button {
+  width: 18.875rem;
+  height: 4.07rem;
+  font-size: 2.25rem;
+  padding: 0;
+  margin: 2rem 0;
+}
```
ZakkMan: Same here re overwriting the styling of another component

Pipfile
```diff
@@ -0,0 +1,11 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+
```
ZakkMan: What's this pipfile doing in the root? Do we need it?

client/src/components/OnboardingCredentials/OnboardingCredentials.js
```diff
@@ -0,0 +1,101 @@
...
+          ></Button>
+        </div>
+      </form>
+    </div>
+  );
+}
+
+OnboardingCredentials.defaultProps = {
+  classList: "OnboardingCredentials"
+};
```
ZakkMan: Do we need this? It doesn't look like you're using props

client/src/components/OnboardingCredentials/OnboardingCredentials.js
```diff
@@ -0,0 +1,101 @@
...
+// import "./OnboardingOne.css";
+// import InputTextField from "../InputTextField/InputTextField.js";
+// import Button from "../Button/Button.js";
+// import OnboardingParent from "../OnboardingParent/OnboardingParent";
+
+// class OnboardingOne extends Component {
+//   constructor(props) {
+//     super(props);
+//     console.log(props);
+
```
ZakkMan: Can you delete this commented out code? We never want to commit commented out code like this

client/src/components/OnboardingCredentials/OnboardingCredentials.js
```diff
@@ -0,0 +1,101 @@
...
+import InputTextField from "../InputTextField/InputTextField.js";
+import Button from "../Button/Button.js";
+
+function OnboardingCredentials(props) {
+  return (
+    <div className={props.classList}>
+      <form className="OnboardingCredentials__form" action="">
+        <h1 className="OnboardingCredentials__header">Step One</h1>
+        <div className="OnboardingCredentials__container">
+          <InputTextField type="text" name="name" placeholder="Name"></InputTextField>
```
ZakkMan: How will we pass the values of these inputs back up? Do we need callbacks or something?

PR #284, opened 2019-12-12
Navbar molecular component PR

wchrapcynski:
A new version of this is in this branch. It'd be great to get a review on this as it is right now since it has so much going on.


client/src/components/Navbar/Navbar.stories.js
```diff
@@ -1,9 +1,17 @@
...
   .add("Default", () => (
-    <Navbar />
-  )
-)
+    <Navbar profileImage={ProfileImage} burgerClick={handleClick} showCondensedMenu={false} />
+  ))
+  // Breakpoint must be achieved for this to fully deisplay (max-width: 992px)
+  .add("Condensed Menu", () => (
+    <Navbar profileImage={ProfileImage} burgerClick={handleClick} showCondensedMenu={true} />
+  ));
```
ZakkMan: I think we only need 1 story for this

client/src/components/Navbar/Navbar.js
```diff
@@ -1,21 +1,67 @@
-import React from 'react';
+import React from "react";
 import "./Navbar.css";
-import Logo from "../Logo/Logo"
+import Logo from "../Logo/Logo";
+import Burger from "../Burger/Burger";
+import ProfilePicture from "../ProfilePicture/ProfilePicture";
+import NavLink from "../NavLink/NavLink";
+import { checkPropTypes } from "prop-types";
```
ZakkMan: It doesn't look like you're using prop types. Can you remove this?

wchrapcynski:
@ZakkMan The functionality of what the menu needs to be when the hamburger icon is clicked isn't in the mock. I have currently have issue #285 to get this created but I need some guidance on what we should make it look like. I don't think the current Navlink component used in the normal Navbar will work in this case. 

With the responsiveness when the hamburger icon is showed, is it the logo in the middle that's not correctly centered as the width of the screen changes? I have been struggling with that a bit. Any suggestions?

client/src/components/Navbar/Navbar.js
wchrapcynski: Ended up actually using them.

wchrapcynski:
Added in condensed Navbar that can display when the burger is clicked.

client/src/components/Navbar/Navbar.js
```diff
@@ -1,21 +1,102 @@
...
   );
-}
+};
+
+Navbar.propTypes = {
+  showCondensedMenu: PropTypes.bool,
+  menu1: PropTypes.string,
+  menu2: PropTypes.string,
+  menu3: PropTypes.string,
+  menu4: PropTypes.string,
```
ZakkMan: @wchrapcynski Looks like this should be an array!

client/src/components/Navbar/Navbar.js
```diff
@@ -1,21 +1,102 @@
...
+            <NavLink
+              text="Book A Trip"
+              to="/"
+              menuList={[
+                { text: "Link 1", to: "/" },
+                { text: "Link 2", to: "/" }
+              ]}
+            />
+            <div className={`Navbar__triangle ${props.menu4}`}></div>
+          </div>
```
ZakkMan: Can you make this a list?

client/src/components/Navbar/Navbar.js
```diff
@@ -1,21 +1,102 @@
...
-import Logo from "../Logo/Logo"
+import Logo from "../Logo/Logo";
+import Burger from "../Burger/Burger";
+import ProfilePicture from "../ProfilePicture/ProfilePicture";
+import NavLink from "../NavLink/NavLink";
+import Icon from "../Icon/Icon";
+import PropTypes from "prop-types";
 
 // Function based React Component
-const Navbar = (props) => {
```
ZakkMan: This component will probably need state (like to track whether the hamburger is open or closed)

client/src/components/Navbar/Navbar.js
ZakkMan: I actually don't see where you're using these props in the component. Could they be removed entirely?

client/src/components/Navbar/Navbar.stories.js
```diff
@@ -1,9 +1,20 @@
...
+function handleClick() {
+  console.log("clicked");
+}
+
+storiesOf("Navbar", module).add("Default", () => (
+  <Navbar
+    profileImage={ProfileImage}
+    searchClick={handleClick}
+    messageClick={handleClick}
+    burgerClick={handleClick}
```
ZakkMan: Can you think of more semantic prop names for these?

client/src/components/Navbar/Navbar.stories.js
```diff
@@ -1,9 +1,20 @@
...
+    showCondensedMenu={true}
+    condensedMenu="Navbar__show"
+    menu2="Navbar__show"
```
ZakkMan: Why do we need to pass classes in like this?

client/src/components/Navbar/Navbar.js
wchrapcynski: @ZakkMan My thought on this is that only one of these is active at any point. So, there will only be one prop being passed down at any given time. Adding the Navbar__show class makes the triangle visible.

client/src/components/Navbar/Navbar.js
wchrapcynski: They were just there to show that the dropdown menu works. Nothing drops down if that array is empty. So, they can be removed.

client/src/components/Navbar/Navbar.js
wchrapcynski: @ZakkMan I originally had it there when I started this component but you had me remove it for some reason.

client/src/components/Navbar/Navbar.stories.js
wchrapcynski: @ZakkMan I felt that they were fairly semantic since they relate to the component that will be activated by them. The burger click could probably be renamed to maybe dropMenuClick. Any suggestions would be great!

client/src/components/Navbar/Navbar.stories.js
wchrapcynski: @ZakkMan I was doing this because this component didn't have state and I was setting it up so that something could be passed down to display condensed menu as well as the triangle for which page is active. If this component should have state, that functionality can be changed.

client/src/components/Navbar/Navbar.js
wchrapcynski: This is now an array where the triangle will be active depending on which one has "Navbar__show" in it.

client/src/components/Navbar/Navbar.js
```diff
@@ -1,21 +1,102 @@
...
+        </div>
+      </div>
+      <div className="Navbar__burger">
+        <Burger onClick={props.burgerClick} active={props.showCondensedMenu} />
+        <div className={`Navbar__condensed-menu ${props.condensedMenu}`}>
+          <NavLink text="My Directory" to="/" menuList={[]} />
+          <NavLink text="Community" to="/" menuList={[]} />
+          <NavLink text="Forum" to="/" menuList={[]} />
+          <NavLink text="Book A Trip" to="/" menuList={[]} />
+        </div>
```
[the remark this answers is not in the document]
wchrapcynski: This was fixed after converting the component to a class and creating a function to create the menu items.

client/src/components/Navbar/Navbar.stories.js
wchrapcynski: This has been changed with the component now being a class and is no longer being passed as a prop.

client/src/components/Navbar/Navbar.js
```diff
@@ -1,21 +1,114 @@
...
-  // Default Class to apply to Component
-  let classList = `Navbar`;
+  // Handles activating the hamburger animation and displays the menu.
+  handleCondensedMenuClick = () => {
+    this.setState({ condensedMenuActive: !this.state.condensedMenuActive });
+    if (this.state.condensedMenuActive === false) {
+      this.setState({ condensedMenu: "Navbar__show" });
+    } else {
+      this.setState({ condensedMenu: "" });
+    }
```
ZakkMan: This should be in your `render` method

client/src/components/Navbar/Navbar.js
```diff
@@ -1,21 +1,114 @@
...
+      // pageMark controls the triangle that marks which page the user is on.
+      // Passing 0-3 as a prop into this component controls where it displays.
+      pageMark: ["", "", "", ""],
+      // menuItems are careated in order of the following array.
+      menuItems: ["My Directory", "Community", "Forum", "Book a Trip"],
+      // The link end points for each menu item.
+      menuLinks: ["/", "/", "/", "/"],
+      // Each item in this array is a set for the dropdown on each menu link.
+      // This is an example for one drop menu item: { text: 'Explore People', to: '/' }
+      menuDropDown: [[], [], [], []]
```
ZakkMan: This data may be static, so we may be able to get away with just having it in `render`

client/src/components/Navbar/Navbar.js
```diff
@@ -1,21 +1,114 @@
...
+  setPageMarker = () => {
+    let setPage = [];
+    for (let i = 0; i < 4; i++) {
+      if (i === this.props.page) {
+        setPage.push("Navbar__show");
+      } else {
+        setPage.push("");
+      }
+    }
+    this.setState({ pageMark: setPage });
```
ZakkMan: This should also be done in render

client/src/components/Navbar/Navbar.js
```diff
@@ -1,21 +1,114 @@
...
+        </div>
+      ));
+    } else {
+      return this.state.menuItems.map((item, key) => (
+        <div className="Navbar__link-item">
+          <NavLink text={item} to="/" menuList={[]} />
+        </div>
+      ));
+    }
+  }
```
ZakkMan: This should be done in render too

client/src/components/Navbar/Navbar.js
```diff
@@ -1,21 +1,114 @@
...
+
+  clickHandler = () => {
+    console.log("Clicked!")
+  }
+
+  componentWillMount() {
+    this.setPageMarker();
+  }
```
ZakkMan: I don't think we need this

client/src/components/Navbar/Navbar.js
```diff
@@ -1,21 +1,114 @@
...
+            <ProfilePicture type="extra-small" link="#" image={this.props.profileImage} />
+          </div>
+        </div>
+        <div className="Navbar__burger">
+          <Burger onClick={this.handleCondensedMenuClick} active={this.state.condensedMenuActive} />
+          <div className={`Navbar__condensed-menu ${this.state.condensedMenu}`}>
+            {this.setMenu()}
+          </div>
+        </div>
       </div>
```
ZakkMan: Can you make it so `Navbar__main` is responsive, rather than having two version of the menu (`Navbar__main` and `Navbar__burger`)?

`Navbar__main` should look like the mock on laptop and desktop and use the hamburger menu and open and close on mobile and tablets.

Let's block off some time today for me to walk you through what I mean

wchrapcynski: I changed it so that the burger menu only shows on mobile. I want talk to you on the best way to get the text areas of the menu to be responsive. I have some ideas, though.

client/src/components/Navbar/Navbar.js
wchrapcynski: Let's talk about this one to clarify things.

client/src/components/Navbar/Navbar.js
wchrapcynski: When I have this sort of code changing state in render, I get errors like this:
```Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.```

File: client/src/components/Navbar/Navbar.js (hunk 11 above)
Comment (wchrapcynski) [reply client/src/components/Navbar/Navbar.js]: I also get the ```Maximum update depth exceeded.``` error when I do this.

File: client/src/components/Navbar/Navbar.js (hunk 12 above)
Comment (wchrapcynski) [reply client/src/components/Navbar/Navbar.js]: This function is being called in render. I did it this way because it is called in more than once place to create the menu for both versions.

File: client/src/components/Navbar/Navbar.js (hunk 11 above)
Comment (wchrapcynski) [reply client/src/components/Navbar/Navbar.js]: Due to how I had to handle some things with the triangle, I really do feel that I need this function. It's now needed for the click handler on the burger icon to help manage displaying the triangle.

File: client/src/components/Navbar/Navbar.js (hunk 10 above)
Comment (ZakkMan) [reply client/src/components/Navbar/Navbar.js]: Did you miss this one?

File: client/src/components/Navbar/Navbar.js (hunk 13 above)
Comment (ZakkMan) [reply client/src/components/Navbar/Navbar.js]: Did you miss this one?

File: client/src/components/Navbar/Navbar.js [hunk 15]
```diff
@@ -1,21 +1,105 @@
...
+  handleCondensedMenuClick = () => {
+    this.setState({ condensedMenuActive: !this.state.condensedMenuActive })
+    if (this.state.condensedMenuActive === false) {
+      this.setState({ condensedMenu: "Navbar__show" });
+      this.setState({ pageMark: ["", "", "", ""] });
+    } else {
+      this.setState({ condensedMenu: "" });
+      this.setPageMarker();
+    }
+  }
```
ZakkMan: Can you simplify this?

ZakkMan: One way would be to use the `.Navbar__show` class - if `condensedMenuActive` is `true`, then add it in the `className` attribute below using a ternary, otherwise don't add it. In your CSS, if the class is there, then the menu is visible, otherwise it isn't.

ZakkMan: Something to be careful of though is calling `setState` twice back-to-back like that. `setState` is asynchronous and this will trigger two renders. Can you merge the two `setState` calls?

client/src/components/Navbar/Navbar.js
```diff
@@ -1,21 +1,105 @@
...
+    };
+  }
 
-  // Default Class to apply to Component
-  let classList = `Navbar`;
+  // Handles activating the hamburger animation and displays the menu.
+  handleCondensedMenuClick = () => {
+    this.setState({ condensedMenuActive: !this.state.condensedMenuActive })
+    if (this.state.condensedMenuActive === false) {
+      this.setState({ condensedMenu: "Navbar__show" });
```
ZakkMan: This toggling the class should just be in the `className` attribute, without storing anything in state.

client/src/components/Navbar/Navbar.js
ZakkMan: Could the nav bar component accept a prop (i.e. from React Router) for the current path and then use that to determine which nav link is "active"? That would make this component much simpler

wchrapcynski: @ZakkMan I have tried putting this in several places in render and the triangle just doesn't display when I do it that way. I was able to get everything else to be in the render section, though.

client/src/components/Navbar/Navbar.js
wchrapcynski: See above.

client/src/components/Navbar/Navbar.js
wchrapcynski: I guess I didn't see the point of changing it yet since we don't have any of the links to put in but I can remove it.

client/src/components/Navbar/Navbar.js
wchrapcynski: > Could the nav bar component accept a prop (i.e. from React Router) for the current path and then use that to determine which nav link is "active"? That would make this component much simpler

I missed this note. I'll look into that.

client/src/components/Navbar/Navbar.js
wchrapcynski: Indeed! I already caught this one, actually, when I moved the function down to the render area this afternoon. I kept working on things after I sent it back to you.

client/src/components/Navbar/Navbar.js
wchrapcynski: I guess I'm not following you on this one. Can you elaborate?

client/src/components/Navbar/Navbar.js
wchrapcynski: @ZakkMan I'm only seeing how to do this if <Route> is calling to render that particular component and using the component attribute.  Route isn't creating the Navbar, right? I'm not sure if that's what you're getting at, though. Right now I have a prop being passed into Navbar that tells it which page is active. I'm using that to put a class in in one spot of an array which gets iterated through to one of the navlinks. I can also make the each array element blank if I want to triangle not to display in the case of the burger menu. I'm not entirely sure how to simplify it. I've been researching it a lot but I can keep looking.

client/src/components/Navbar/Navbar.js
wchrapcynski: I figured out how to do this with ```<div className={`Navbar__main ${this.state.condensedMenuActive && `Navbar__show`}`}>```

client/src/components/Navbar/Navbar.js
wchrapcynski: This whole if/else statement is gone now.

client/src/components/Navbar/Navbar.js
wchrapcynski: This is the only one that I can't figure out how to get rid of at this point. I have now condensed everything else.

wchrapcynski:
> This all looks good, but I get this error when I try to run it:
> 
> ```
> RROR in ./src/components/Navbar/Navbar.stories.js
> Module not found: Error: Can't resolve '../../assets/images/profile_large.png' in '/Users/zakk/projects/CoTrip/client/src/components/Navbar'
> ```
> 
> Can you use the alias that Blake setup for the images?

This is actually from the default profile image in the Navbar. This will normally get passed as a prop.   I didn't realize that the profile pictures were changed in the assets folder this morning which caused this particular error. I went ahead and pointed it to the new file and all looks good now.

client/src/components/Navbar/Navbar.js
```diff
@@ -1,21 +1,73 @@
...
+                {this.props.page === key && (
+                  <div className={`Navbar__triangle`}></div>
+                )}
+              </div>
+            ))}
+          </div>
+          <div className="Navbar__right">
+            <Icon icon={"search"} size="2x" onClick={this.clickHandler} />
+            <Icon icon={["far", "comment-dots"]} size="2x" onClick={this.clickHandler} />
+            <ProfilePicture type="extra-small" to="/" image={this.props.profileImage} />
```
ZakkMan: Just caught this - this should link to the profile page for the current user. Could we have this be a prop that gets passed down?

wchrapcynski: Done. Added new prop to be passed called profileUrl.

client/src/components/Navbar/Navbar.js
```diff
@@ -1,21 +1,75 @@
...
+Navbar.propTypes = {
+  page: PropTypes.number,
+  image: PropTypes.string,
+  to: PropTypes.string
+};
+
+Navbar.defaultProps = {
+  page: 0,
+  image: "../../assets/images/profile-picture-1.png",
+  to: "/"
```
ZakkMan: should this be `profileUrl`?

wchrapcynski: Done. It's now to: this.props.profileUrl

wchrapcynski: Well, I just noticed that broke it. Reverting back.

ZakkMan: Okay - let me know when you fix it. Once you've fixed it, then send it back for another review

wchrapcynski: Moved default props to be inside of render and that seems to work.

Issue #285, opened 2019-12-12
Condensed popup Navbar menu Issue


wchrapcynski:
This is the menu that will display when the burger icon is clicked. The burger icon is displayed when the website not wide enough to show the entire main menu. There is nothing in the mock designed for this currently. @ZakkMan Can you advise on how to proceed?

ZakkMan:
@wchrapcynski Can you come up with a design, using the elements we have? We'll want a variation on the navlinks that we have, but in a takeover-style menu

wchrapcynski:
I'm not entirely sure how well it will work out but I'll make an attempt.

PR #298, opened 2019-12-13
Component navlink condensed PR

wchrapcynski:
Per issue #285, this is the initial design for the menu that would come up with the burger icon is clicked. Since there wasn't a design for this in the mock, I went and and started one. I'm not sure if it needs to be more complicated than this. Let me know what you think.

### DIFF
--- a/client/src/components/NavlinkCondensed/NavlinkCondensed.css
+++ b/client/src/components/NavlinkCondensed/NavlinkCondensed.css
@@ -1,0 +1,4 @@
+
+.NavlinkCondensed {
+
+}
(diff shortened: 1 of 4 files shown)


ZakkMan:
@wchrapcynski Rather than have this be a separate component, it should be a responsive variation on the existing navbar component

wchrapcynski:
@ZakkMan It is. The component that displays the Navlinks, in my opinion, does not work what's needed to display when the burger is clicked. I think the Navlink component would have to be modified or we need another component.

ZakkMan:
@wchrapcynski Can you update the existing components? for the navbar and navbar links? Those should be all we need to implement the full navbar, including responsive.

Issue #299, opened 2019-12-13
Molecular Component: OnboardingPills


stuvun:
![onboardingPills](https://user-images.githubusercontent.com/55634628/70816307-34bb9100-1d9d-11ea-8b85-95619daf1899.png)

This is using the Pill component and displaying a list of them for the Onboarding 2 page.

PR #300, opened 2019-12-13
Component onboarding pills

stuvun:
#299


client/src/components/OnboardingPills/OnboardingPills.js
```diff
@@ -0,0 +1,27 @@
...
+      />
+    );
+  });
+  return (
+    <div className={classList}>
+      <div className="OnboardingPills__inner-container">{renderPills}</div>
+    </div>
+  );
+};
+
```
llucas314: Looks really good, I would just add in a default prop array for the pills prop.

client/src/components/OnboardingPills/OnboardingPills.css
```diff
@@ -0,0 +1,19 @@
...
+.OnboardingPills .Pill {
+  width: 9.75rem;
+  height: 2.45rem;
+  font-size: 1.25rem;
+  margin: 0.5rem;
+  padding: 0;
+  box-shadow: none;
+  line-height: 2;
+  user-select: none;
+}
```
ZakkMan: All this CSS should be part of the `Pill` component - if you need to change something, then you should make the changes there!

client/src/components/OnboardingPills/OnboardingPills.css
```diff
@@ -0,0 +1,19 @@
+.OnboardingPills {
+  display: inline-block;
+  width: 34.6rem;
+}
+.OnboardingPills__inner-container {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: center;
+}
```
ZakkMan: Are you maybe able to merge these two things?

PR #301, opened 2019-12-13
Comment body component PR

karenheyn:
Changed component from % to rem,
added media queries


client/src/components/CommentBody/CommentBody.css
```diff
@@ -1,19 +1,66 @@
+body {
+  text-align: center;
+}
```
ZakkMan: We don't want to have a global selector like this because it may clash with styles in other components

client/src/components/CommentBody/CommentBody.css
```diff
@@ -1,19 +1,66 @@
...
+
 .CommentBody {
```
ZakkMan: I think this class name should be `Comment__body`, per the naming convention in the readme

client/src/components/CommentBody/CommentBody.css
```diff
@@ -1,19 +1,66 @@
...
   height: auto;
   background-color: var(--gray-dark);
   font-size: 14px;
   font-family: var(--paragraph-font);
+  padding: 1.2rem;
+}
+
+.CommentBody p {
+  text-align: left;
+}
```
ZakkMan: Can you think of a class name that's inline with the naming convention?

client/src/components/CommentBody/CommentBody.css
```diff
@@ -1,19 +1,66 @@
...
+  height: 90%;
+  margin-bottom: 0;
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+  height: 9rem;
+}
+
+.CommentBody_center_item {
+}
```
ZakkMan: If this is empty, can we delete it?

client/src/components/CommentBody/CommentBody.css
```diff
@@ -1,19 +1,66 @@
...
+
+.CommentBody_name {
+  text-align: center !important;
```
ZakkMan: Ah! Please no `!important`

client/src/components/CommentBody/CommentBody.css
```diff
@@ -1,19 +1,66 @@
...
+.CommentBody_activity {
+  text-align: left;
+  margin-right: 0.8rem;
+  margin-top: 0.5rem;
+  display: flex;
 }
-.CommentBody__date {
-  margin-left: 10px;
+
+.Date {
```
ZakkMan: `.Date` is a pretty greedy selector (meaning it could end up applying to a lot of places. Can you do something more inline with the naming convention? `.CommentBody__date` is good (the one deleted)

client/src/components/CommentBody/CommentBody.js
```diff
@@ -1,22 +1,33 @@
...
 
   return (
     <div className="CommentBody">
-      <p>{body}</p>
-      <div className="CommentBody__date">
+      <span className="CommentBody_item">
+        <ProfilePicture type={type} to={to} image={image} />
+        <p className="CommentBody_name">{name}</p>
+      </span>
+      <span className="CommentBody_item CommentBody_center_item">
```
ZakkMan: All `span` huh? Could you come up with some more semantic HTML to use?

client/src/components/Comment/Comment.css
```diff
@@ -0,0 +1,67 @@
...
+  align-items: flex-start;
+  width: 50rem;
+  height: auto;
+  background-color: var(--gray-dark);
+  font-size: 14px;
+  font-family: var(--paragraph-font);
+  padding: 1.2rem;
+}
+
+.Comment_h4 {
```
ZakkMan: Can you come up with a name for this, rather than using the tag name? We want the name to be semantic

client/src/components/Comment/Comment.css
```diff
@@ -0,0 +1,67 @@
...
+  margin: 0;
+}
+
+.Comment_section {
```
ZakkMan: Same here - a semantic name

client/src/components/Comment/Comment.css
```diff
@@ -0,0 +1,67 @@
...
+  margin: 0.7rem;
+  height: 90%;
+  margin-bottom: 0;
+}
+
+.Comment_section_1 {
```
ZakkMan: Same here - a semantic name. Also, you never want to put numbers in a class name. This is where you'd use modifiers on a block(re BEM in the readme)

ZakkMan: Also, can you make these classnames in line with the naming convention?

ZakkMan: Two underscores, two dashes, etc

client/src/components/Comment/Comment.css
```diff
@@ -0,0 +1,67 @@
...
+.Comment_section_3 {
+  white-space: nowrap;
+  text-align: right;
+}
+
+.Comment_section_3_p {
+  margin: 0;
+}
+
+.Comment_div {
```
ZakkMan: Same here - a semantic name

PR #325, opened 2019-12-16
Component: Card, (added mobile cards) PR

TylerEikenberg:
Realized we needed these cards for the mobile site so I added an extra size class to the css for the card component, and extra colors. 
Screen Shot 2019-12-16 at 10 41 19 AM


client/src/components/Card/Card.css
```diff
@@ -43,6 +48,26 @@
   min-width: 180px;
```
erinkelley27: You have two `min-width` properties here.

client/src/components/MediaCard/MediaCard.css
```diff
@@ -1,9 +1,13 @@
 .MediaCard {
```
erinkelley27: Is the MediaCard needed for your new cards at all?

TylerEikenberg: @erinkelley27  No! Not needed.

PR #334, opened 2019-12-16
User Auth Tokens and Profile Model PR

torykling:
Paths seem to be working when we tested them. 

We can register a new user, login to get a token, create a profile, read all profiles, and read, update, and delete a specific profile. 

We wanted to make a pr at this point so we can start testing log in on the front end while other members of the backend team continue to work on other models (trips, topics, etc)


api/Pipfile
```diff
@@ -10,6 +10,10 @@ pylint = "*"
 django = "*"
 psycopg2-binary = "*"
 autopep8 = "*"
+djangorestframework-simplejwt = "*"
```
ZakkMan: Are you still using this? If not, can we remove it?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.pythonPath": "/Users/victoriaklingenstein/.local/share/virtualenvs/CoTrip-TXuPc5lv/bin/python"
+}
```
ZakkMan: Can you delete these settings files?

api/accounts/urls.py
```diff
@@ -0,0 +1,15 @@
...
+from .views import ProfileList, ProfileDetail
+
+urlpatterns = [
+    path('api/auth', include('knox.urls')),
+    path('api/auth/register', RegisterAPI.as_view()),
+    path('api/auth/login', LoginAPI.as_view()),
+    path('api/auth/user', UserAPI.as_view()),
+    path('api/auth/logout', knox_views.LogoutView.as_view(), name='knox_logout'),
+    path('api/profile', ProfileList.as_view(), name='profile_list'),
+    path('api/profile/<int:pk>', ProfileDetail.as_view(), name='profile_detail')
```
ZakkMan: Maybe the `api` part of these urls is unnecessary?

.vscode/settings.json
torykling: Deleted. Not sure where these came from.

api/Pipfile
torykling: Uninstalled.

api/accounts/urls.py
torykling: Removed them

PR #349, opened 2019-12-17
made model, view, and serializer for trips

jennadean9:
planning on linking profile to trips on the profile model, hopefully tomorrow and keep it at a one to many (person to trips)


api/api/settings.py
```diff
@@ -37,7 +37,9 @@
     'django.contrib.sessions',
     'django.contrib.messages',
     'django.contrib.staticfiles',
-    "cotrip_app",
+    'cotrip_app',
```
ZakkMan: What is the `cotrip_app`? Do we need it?

api/api/urls.py
```diff
@@ -15,7 +15,9 @@
 """
 from django.contrib import admin
 from django.urls import path
+from django.conf.urls import include
 
 urlpatterns = [
     path('admin/', admin.site.urls),
+    path('', include('trip.urls'))
```
ZakkMan: Should these be namespaced under `trip` or something?

api/trip/models.py
```diff
@@ -0,0 +1,27 @@
+from django.contrib.postgres.fields import ArrayField
+from django.db import models
+import sys
+sys.path.append("..")
```
ZakkMan: What is this for?

api/trip/models.py
```diff
@@ -0,0 +1,27 @@
+from django.contrib.postgres.fields import ArrayField
```
ZakkMan: Doesn't look like you're using an `ArrayField` - can we remove this?

api/trip/models.py
```diff
@@ -0,0 +1,27 @@
...
+from django.db import models
+import sys
+sys.path.append("..")
+
+
+class Trip(models.Model):
```
ZakkMan: These models look great

api/trip/serializers.py
```diff
@@ -0,0 +1,27 @@
+from .models import Trip, Activity
+from rest_framework import serializers
+
+
+class TripSerializer(serializers.HyperlinkedModelSerializer):
+    activities = serializers.HyperlinkedRelatedField(
+        view_name='activity_detail',
+        many=True,
+        read_only=True
+    )
```
ZakkMan: It'll be easier for the frontend if this is a [nested relationship](https://www.django-rest-framework.org/api-guide/relations/#nested-relationships)

api/trip/urls.py
```diff
@@ -0,0 +1,19 @@
+from django.urls import path
+from . import views
+
+urlpatterns = [
```
ZakkMan: Could you arrange these so they're grouped by trip and activity?

ZakkMan: Also, they should be singular (i.e. `trip`, `activity`)

api/trip/views.py
```diff
@@ -0,0 +1,25 @@
+from rest_framework import generics
+from .serializers import TripSerializer, ActivitySerializer
+from .models import Trip, Activity
+from django.views.generic import ListView, DetailView
+from django.views.generic.edit import CreateView, DeleteView, UpdateView
+
+
+class TripList(generics.ListCreateAPIView):
```
ZakkMan: Can you work with the Auth team and add permissions here? We only want admins to be able to create trips and activities

api/api/settings.py
jennadean9: it is where the location and city are going to go, but ill delete it on this PR!!! se can merge when those get a PR

Pipfile
```diff
@@ -0,0 +1,11 @@
...
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+
+[packages]
+
+[requires]
+python_version = "3.7"
```
ZakkMan: What is this? Do we need it? I think we can delete it

api/api/settings.py
```diff
@@ -36,8 +36,8 @@
     'django.contrib.contenttypes',
     'django.contrib.sessions',
     'django.contrib.messages',
-    'django.contrib.staticfiles',
```
ZakkMan: Rut roh! What happened to `staticfiles`? I mean, if we don't need it then it's cool

ZakkMan: Looks like we need `staticfiles`:

Screen Shot 2019-12-17 at 3 12 36 PM

.gitignore
```diff
@@ -401,4 +401,7 @@ $RECYCLE.BIN/
 # Jest Testing Snapshots
 browser/src/__snapshots__
 
+# Images folder
+api/images
```
ZakkMan: Much better - you may have to delete the folder, since it's still there with all the images uploaded to it

api/trip/urls.py
```diff
@@ -0,0 +1,20 @@
...
+    path('', views.TripList.as_view(), name='trip_list'),
+    path('<int:pk>', views.TripDetail.as_view(), name='trip_detail'),
+    path('new', views.TripList.as_view(), name='trips_create'),
```
ZakkMan: I get this error when I try to create a new trip:

```sh
AssertionError at /trip/
The `.create()` method does not support writable nested fields by default.
Write an explicit `.create()` method for serializer `trip.serializers.TripSerializer`, or set `read_only=True` on nested serializer fields.
```

Pipfile
```diff
@@ -0,0 +1,11 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+
+[packages]
+
+[requires]
```
ZakkMan: Can you delete this file?

api/Pipfile
```diff
@@ -10,10 +10,14 @@ pylint = "*"
...
 psycopg2-binary = "*"
 autopep8 = "*"
+<<<<<<< HEAD
+djangorestframework = "*"
+=======
 pillow = "*"
 djangorestframework = "*"
 django-rest-knox = "*"
 django-cors-headers = "*"
+>>>>>>> develop
```
ZakkMan: Looks like a merge was committed! Can you fix this?

Issue #350, opened 2019-12-17
Component: Group Card Bug Issue


shelbyvjacobs:
From Issue #176.

The group card needs to be updated for two reasons:
1. Image props are not being passed in correctly
2. If the name of the group is too long, it pushes the # of members and location off of the card.

caioingber:
Issue confirmed resolved, closing

PR #357, opened 2019-12-17
Component: Group Card Bug PR

shelbyvjacobs:
From Issue #350.

I fixed the bug with the background props not displaying properly. I spent the morning trying to make the styling work so that the name could be two lines, but I couldn't find a solution. I had other people look at it as well, but we were stumped. There's something that is making the positioning really difficult! I talked to Josh, and we decided that it isn't the most important thing I could be working on right now. I will return to that part of the issue when I have time. In the meantime, do you guys have any suggestions?


shelbyvjacobs:
You're a GENIUS! I can't believe I didn't think of using flexbox there too... usually I'm so excited to use flexbox and it's the first thing I try! Thanks!

PR #361, opened 2019-12-17
Component/Page: Coming Soon PR

TylerEikenberg:
I made a page that just displays the CoTripper yellow and the words "Coming Soon". I made this for all the buttons on pages that don't lead anywhere, I figured we could have them link to this page instead of nothing. Totally up to you guys if you want to include this, I just thought it might be cool!

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -1,20 +1,47 @@
 import React, { Component } from "react";
 import { Route } from "react-router-dom";
 import SplashPage from "./pages/SplashPage/SplashPage";
+import ComingSoonPage from "./pages/ComingSoonPage/ComingSoonPage";
 import "./App.css";
-import { library } from '@fortawesome/fontawesome-svg-core'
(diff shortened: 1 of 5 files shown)


client/src/App.js
```diff
@@ -1,20 +1,47 @@
...
+);
 
 class App extends Component {
   render() {
     return (
-      <div className='App'>
+      <div className="App">
         <main>
           <Route path="/" exact component={SplashPage}></Route>
+          <Route path="/comingsoon" exact component={ComingSoonPage}></Route>
```
ZakkMan: Could we have the path be `coming-soon`?

ZakkMan: or `coming_soon`

TylerEikenberg: Lol, I'll fix it, thanks 👍

TylerEikenberg:
@ZakkMan added your suggestions, let me know what ya think

PR #362, opened 2019-12-17
Component: Book A Trip Page Images PR

shelbyvjacobs:
I adjusted the image size for the largest screen size. It storybook, it looks too large, but in the actual page this is the right size. I tested in the page to make sure. Also, I added another level of responsiveness using a media query.


client/src/components/BookATripImage/BookATripImage.js
```diff
@@ -10,10 +10,10 @@ const BookATripImage = (props) => {
       <div className="BookATripImage__image-group">
         {cards.map(card => {
           return (
-            <div className="BookATripImage__image-container">
-              <img src={require(`../../assets/images/${card.image}.png`)} className="BookATripImage__image"></img>
-              <div><Link to={card.url} className="BookATripImage__link">{card.description}</Link></div>
-            </div>
+              <div className="BookATripImage__image-container">
+                <img src={require(`../../assets/images/${card.image}.png`)} className="BookATripImage__image"></img>
```
ZakkMan: This is very clever! However, these images may come from the backend at some point. So it'd be better to accept the image as a prop

shelbyvjacobs:
Okay! Everything is working now! Yay!

PR #367, opened 2019-12-17
Component banner image card

cenwachukwu:
Created the large banner-like image card component. I also made it responsive


client/src/components/BannerImageCard/BannerImageCard.css
```diff
@@ -0,0 +1,46 @@
...
+        margin: 0 auto;
+        margin-top: 2rem;
+        display: flex;
+        flex-direction: column;
+        justify-content: center;
+        align-content: center;
+        text-align: center;
+        font-size: 3rem;
+    }
+}
```
erinkelley27: Remove the styles here that are repeated from the other media query. They are only needed in one place - when you use max-width, the style apply to the specified screen size and smaller.

client/src/components/BannerImageCard/BannerImageCard.stories.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import BannerImageCard from './BannerImageCard';
+import Banner__Image from "./TripBanner.png";
```
erinkelley27: Add this to the assets directory instead of this component's directory.

Issue #347, opened 2019-12-16
Component: Slider Component Responsiveness Issue


shelbyvjacobs:
From Issue #79 and PR #199. 

![Screen Shot 2019-12-16 at 4 24 46 PM](https://user-images.githubusercontent.com/53311285/70944290-c4bd3d00-2020-11ea-8217-31364fffc487.png)

It seems that regardless of the screen size, the overview is always below the slider! It should be to the right unless the viewport is small, like here: 

![image](https://user-images.githubusercontent.com/53311285/70944362-e9b1b000-2020-11ea-9eae-a43a58617cb0.png)

Ideally Tory (who originally worked on this component) or the Responsiveness team will take this issue. It's needed for the Hawaii 2020, which is a silver level page.

PR #369, opened 2019-12-18
Slider container (New issue #347)

stuvun:
#347


client/src/components/TripDetails/TripDetails.js
```diff
@@ -37,18 +37,24 @@ class TripDetails extends Component {
           </div>
         </div>
         <div className="TripDetails__right">
```
erinkelley27: Change the class name here since this div is not always on the right.

client/src/components/TripDetails/TripDetails.css
```diff
@@ -1,108 +1,119 @@
...
+  .TripDetails__title {
+    font-size: 2.25rem;
+    min-width: 40%;
+  }
+  .TripDetails__text {
+    display: flex;
+    flex-wrap: wrap;
+    max-width: 60%;
   }
 }
```
erinkelley27: Change height and width percentages to rem where needed. Check in with the responsive team if you need to!

PR #377, opened 2019-12-18
Changed User Creation

bmillsmc:
Changed the view for User creation, so that when a user is created you can input a `profile` field with a profile object and that profile will be made and assigned to the user. Updating the profile is still being worked on, but this merge is just for the creation so front-end can test linking with the backend.


api/accounts/serializers.py
```diff
@@ -1,5 +1,4 @@
 from rest_framework import serializers
-# from django.contrib.auth.models import User
```
ZakkMan: Can you delete any commented out code?

api/accounts/views.py
```diff
@@ -1,17 +1,93 @@
...
+#         "last_name": "J",
+#         "city_of_residence": "Chicago",
+#         "age": 50,
+#         "dream_destination": "Chicago",
+#         "bio": "I hang out with loony tunes",
+#         "instagram_url": "",
+#         "pinterest_url": "",
+#         "facebook_url": ""
+#     }
+# }
```
ZakkMan: This could go in documentation. We only want documenting-style comments in the codebase though

bmillsmc:
I fixed the two issues you had with it. I tested the creation of a user with a profile field, and it works. I'm still working on the view for updating the profile, so that doesn't work yet. Should I delete the update profile view from this branch and make a new feature branch to work on that?

reisrael4:
@ZakkMan, could you give more specific feedback re: which comments still need to be addressed? I looked through each one and nothing jumped out as not having been dealt with.

ZakkMan:
@reisrael4 Can you test this out to make sure it all works? The code looks good and I'm pretty sure they finished this but if you don't mind running through it. make sure you can sign up, sign in, get your profile, etc. I'd test the API through postman and then test the API with the frontend.

reisrael4:
@ZakkMan, I'm starting to think that this issue was resolved on another branch and this was just not closed at any point because it functions beautifully and without errors from develop and accomplishes everything that this branch says above that it's trying to accomplish, but whenever I try and run it from this branch everything is broken.

PR #380, opened 2019-12-18
Connect frontend backend PR

torykling:
I hooked up login page and onboarding with backend and set up redirects for logged in users.

I also added a logout button to the coming soon page. We don't currently have a logout button anywhere in the figma...

The user is successfully getting created at registration, but the profile is not. When Ben's work on the profile model gets merged the profile **should** get created.


torykling:
The profile is working and getting created on registration. Topics are not working. We were getting an error related to the many to many relationship. So we removed topics from the data set getting passed into the post request. The topic pills are clickable but useless.

ZakkMan:
@torykling Does this work with the stuff @bmillsmc is working on? Can y'all send me just 1 PR to review with the whole login and onbaoarding flow?

PR #381, opened 2019-12-18
Responsive Changes, Card PR

TylerEikenberg:
I had to totally revamp the original base card component that I had created to make it responsive which affected all of the molecular card components as a result so I had to go through many of the card components and make various changes to use the new card component changes. Please let me know what you think, I know theres a lot of changes and I'm still working on some of it. If you view the homepage you'll see the card components are now totally responsive. 

Some of the components in storybook are going to look janky but I've been trying to prioritize it by which components are on pages that we're currently working on and will be presenting.


client/src/components/GroupCard/GroupCard.css
```diff
@@ -1,39 +1,80 @@
...
+  }
+  .GroupCard__details {
+  font-size: 1rem;
+  }
+}
+
+/* Medium devices */
+@media only screen and (max-width: 768px) {
+ .GroupCard__group-name {
+    font-size: 1.8vw;
```
ZakkMan: rem or px is probably better for font-size. Can you use one of those?

ZakkMan: If you need to change the value, you can do so in a media query

TylerEikenberg: @ZakkMan There is HomePage stuff in here, I started messing with stuff from my HomePage branch and then decided to make this new branch, but i've accounted for all of the changes in regards to that. I'll take a look at the font size 💯

TylerEikenberg:
@ZakkMan

PR #390, opened 2019-12-19
Member profile page PR

karenheyn:
full page with all components


client/src/components/Bio/Bio.stories.js
```diff
@@ -0,0 +1,31 @@
...
+  <Bio
+    instagram={people[0].intagram}
+    facebook={people[0].facebook}
+    pinterest={people[0].pinterest}
+    type="default"
+    name={people[0].name}
+    bio={people[0].bio}
+    hashtags={people[0].hashtags}
+    isCurrentUser={people[0].isCurrentUser}
+    onClick={console.log("Hello")}
```
ZakkMan: You'll need to make this a function

client/src/components/MemberProfilePage/MemberProfilePage.css
```diff
@@ -0,0 +1,69 @@
+.MemberProfilePage {
```
ZakkMan: Can you fix these classes so they're inline with BEM?

client/src/components/MemberProfilePage/MemberProfilePage.css
```diff
@@ -0,0 +1,69 @@
...
+
+.page {
+  display: flex;
+  max-width: 100vw;
+
+}
+
+.MemberProfilePage .BookTripCard__content-wrapper {
+  width: 90%;
+}
```
ZakkMan: We don't want to change the CSS of other components

ZakkMan: This will make working on that component really difficult

client/src/components/MemberProfilePage/MemberProfilePage.js
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  // Runs after Component is loaded in the broswer
+  componentDidMount() {}
+
+  // Runs after a component has been updated
+  componentDidUpdate() {}
+
+  // Runs right before a component is removed from the DOM
+  componentWillUnmount() {}
```
ZakkMan: If you don't have any state, then could you turn this into a function and delete all this?

client/src/components/MemberProfilePage/MemberProfilePage.js
```diff
@@ -0,0 +1,170 @@
...
+        userSurname: "Knope",
+        userPic: pic3
+      },
+      {
+        userId: 7,
+        userFirstName: "Frank",
+        userSurname: "Ocean",
+        userPic: pic1
+      }
+    ];
```
ZakkMan: Can you pass this data in as props?

client/src/components/MemberProfilePage/MemberProfilePage.js
```diff
@@ -0,0 +1,170 @@
+import React, { Component } from "react";
```
ZakkMan: If this is a page, can you move it to the `pages/` directory?

PR #391, opened 2019-12-19
Book A Trip Page PR

shelbyvjacobs:
This is the draft of the Book a Trip page; changes will still be made to improve formatting, etc. For instance, the text input field in the banner stretches the teeny weeny viewport.


ZakkMan:
@shelbyvjacobs do you think this is finished?

shelbyvjacobs:
I removed the BookATripsPage from components; it's only in pages now.

PR #394, opened 2019-12-19
Hawaii 2020 Page PR

shelbyvjacobs:
Like the Book A Trip Page, this page is 90% done and can be tested, but there are basic formatting issues that need to be addressed.


client/src/App.js
```diff
@@ -3,8 +3,12 @@ import { Route } from "react-router-dom";
 import SplashPage from "./pages/SplashPage/SplashPage";
 import HomePage from "../src/pages/HomePage/HomePage";
 import ComingSoonPage from "./pages/ComingSoonPage/ComingSoonPage";
+<<<<<<< HEAD
```
ZakkMan: Can you fix these merges that were commited?

PR #402, opened 2019-12-19
Card Component styling issue.

stuvun:


ZakkMan:
@stuvun What am I looking at here? Can you provide a description?

stuvun:
So for some reason, the styling I added to the Card component was overwritten by someone else's so the ActivityList component wasn't displaying properly. All the changes were for the "Card--extra-small" class, which is exclusively for the ActivityList component.

client/src/components/Card/Card.css
```diff
@@ -45,9 +45,18 @@
   max-height: 125px;
 }
 .Card--extra-small {
-  width: 1rem;
-  max-width: 185px;
-  max-height: 135px;
+  min-width: 6rem;
+  max-width: 6rem;
+  min-height: 4.32rem;
+  max-height: 4.32rem;
```
ZakkMan: Do you need to do this? Could you just set a height and width?

stuvun: It didn't seem to work out any other way, It might be a padding or wrapping issue, but the code where I had it working is on my desktop, so I can't check right now.

stuvun: Actually let me check my last pull request for the activitylist component

stuvun: Just changed the styling back to what was at my pull request. The only difference I could really find was I didn't set a min-max for the height.

Issue #397, opened 2019-12-19
Burger icon is sticky


wchrapcynski:
The burger icon sticks to the top right corner of the page and stays there when you scroll.

andrew-frazier6:
@ZakkMan Took a break from Storybook Organization to work on this. Could you provide some context into where this burger would go and how many links I will need to create in the dropdown? I have scratched the old setup because of a bug that wasnt allowing it to show up so this will allow me to tailor the new Burger for exactly where and what we will need

caioingber:
@andrew-frazier6 per our convo, this issue has been resolve in PR #403 , addressing additional issues in #285

PR #403, opened 2019-12-19
This should fix the issue with the burger icon sticking - PR

wchrapcynski:
This should take care of the problem in issue #397. I removed the position tag from the burger icon.

### DIFF
--- a/client/src/components/Navbar/Navbar.css
+++ b/client/src/components/Navbar/Navbar.css
@@ -42,7 +42,6 @@
 .Navbar__burger {
   margin: 0 20px;
   display: none;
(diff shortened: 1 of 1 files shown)


client/src/components/Navbar/Navbar.css
```diff
@@ -42,7 +42,7 @@
 .Navbar__burger {
   margin: 0 20px;
   display: none;
-  position: absolute;
+  /* position: absolute; */
```
ZakkMan: Don't comment out this line, delete it

wchrapcynski: Done.

PR #411, opened 2019-12-19
Card Component Quick Fix

stuvun:
Fixed Errors


client/src/components/Card/Card.css
```diff
@@ -46,18 +46,16 @@
 }
 .Card--extra-small {
   color: var(--white);
-  min-width: 180px;
+  min-width: 7rem;
+  max-width: 7rem;
+  min-height: 5rem;
```
ZakkMan: hey, do we need to set explicit min and max widths and heights? How does that hold up with responsive?

ZakkMan: @Mdusseau did you miss this comment?

client/src/components/Card/Card.css
```diff
@@ -12,57 +12,65 @@
...
+
 .Card--yellow {
   background-color: var(--yellow);
 }
+
 .Card--large {
   width: 32vw;
   max-width: 500px;
   min-width: 200px;
   max-height: 275px;
```
ZakkMan: This is a little odd - do we need to set max widths, min-widths, a width, and a max height? We definitely want this to have a minimum size if it's a large card

client/src/components/Card/Card.css
```diff
@@ -12,57 +12,65 @@
...
 }
+
 .Card--medium {
   width: 14.2vw;
   max-width: 225px;
   max-height: 150px;
   min-width: 180px;
```
ZakkMan: Same here re my comment above on the sizing

client/src/components/Card/Card.css
```diff
@@ -12,57 +12,65 @@
...
-  margin: 0 0.1vw 0 0.1vw;
+  margin: 0 0.2vw 0 0.2vw;
```
ZakkMan: Can you use rem instead of vw?

ZakkMan: vw is good when something needs to be sized relative to the window (like I want the header to always take up 30% of the screen). Otherwise, it's always better to use rem

client/src/components/Card/Card.css
```diff
@@ -12,57 +12,65 @@
...
 }
+
 .Card--small {
   max-width: 225px;
   max-height: 125px;
 }
```
ZakkMan: Does this max width hold up if there's a lot of content in the small card?

client/src/components/Card/Card.css
```diff
@@ -12,57 +12,65 @@
...
+
 .Card--extra-small {
   color: var(--white);
-  min-width: 180px;
+  min-width: 7rem;
+  max-width: 7rem;
+  min-height: 5rem;
+  font-size: 0.9rem;
```
ZakkMan: I think there are variables for the font sizes. Can you use that?

client/src/components/Card/Card.css
```diff
@@ -81,20 +89,23 @@
   max-height: 202px;
   min-height: 200px;
 }
```
ZakkMan: Can this go in a media query? Rather than calling the class Card--mobile?

client/src/components/Card/Card.css
```diff
@@ -81,20 +89,23 @@
...
   text-align: center;
-
   padding-bottom: 3vh;
   width: 100%;
 }
+
 .Card--mobile h2 {
   text-align: center;
   width: 100%;
 }
```
ZakkMan: Can you create classnames that are inline with the naming convention? (i.e. `.Card--mobile__title` instead of `.Card--mobile h2`)

client/src/components/Card/Card.css
```diff
@@ -105,24 +116,29 @@
   -moz-box-shadow: none;
   box-shadow: none;
```
ZakkMan: Do wee need these browser prefixes? I don't think we do

client/src/components/Card/Card.css
```diff
@@ -105,24 +116,29 @@
...
 }
+
 .Card__discover-link {
   color: black;
```
ZakkMan: Can you use the CSS variable here?

client/src/components/Card/Card.css
```diff
@@ -105,24 +116,29 @@
...
   text-align: center;
   transition: transform 0.3s;
 }
+
 .Card__discover-link a {
```
ZakkMan: Naming convention here too

ZakkMan: Or maybe this can use the Button component?

client/src/components/Card/Card.css
```diff
@@ -105,24 +116,29 @@
...
   text-decoration: none;
 }
+
 .Card__discover-link:hover {
   transform: scale(1.1);
 }
+
 .Card__discover-link a:hover {
   color: var(--yellow);
 }
```
ZakkMan: This seems like contextual information when we want these to be abstracted. Is there a better way of naming this?

client/src/components/Card/Card.css
Mdusseau: No, I looked in the root and did not see any variables for font sizes.

PR #417, opened 2020-03-30
Escalation templates

caioingber:
Issue: #416 

If I understood the documentation correctly - I created a hidden .github directory at the root level and included a `ISSUE_TEMPLATE` subdirectory with two different templates, as well as a `PULL_REQUEST_TEMPLATE` subdir with just the singular template. Let me know if I need to change the file structure or any other changes you'd like me to make to the templates themselves.

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,9 +1,0 @@
-- Related Issue (include '#'):
-
-- Description of changes made:
-
-- Is the feature complete/bug resolved/etc..:
-
-- Any known bugs/strange behavior:
-
-- Is there specific feedback you would like on these changes:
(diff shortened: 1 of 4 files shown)


pull_request_template.md
```diff
@@ -0,0 +1,9 @@
+- Related Issue (include '#'):
+
+- Description of changes made:
+
+- Is the feature complete/bug resolved/etc..:
+
+- Any known bugs/strange behavior:
+
+- Is there specific feedback you would like on these changes:
```
ZakkMan: @caioingber Do we need this file too? It's in the root of the repository

caioingber: @ZakkMan Hmmm not sure why that was included in the PR i moved that file into the template folder

caioingber: I removed it from the pull request

ZakkMan: Great!

PR #418, opened 2020-03-30
added location models to models, views, serializers, urls

zfriendly:
Created a location model in the cotrip_app folder, then created serializers, and updated the views and urls to reflect the changes. Not sure if city should be a foreign key linked to user profiles, but figures this was a start.


Pipfile
```diff
@@ -0,0 +1,11 @@
+[[source]]
```
ZakkMan: What are this Pipfile and Pipfile.lock? The project already has these

api/cotrip_app/models.py
```diff
@@ -6,3 +6,17 @@ class Topic(models.Model):
...
     def __str__(self):
         return self.title
+
+class Location(models.Model):
+    city = models.CharField(max_length=75)
+    state = models.CharField(max_length=30, null=True, blank=True)
+    country = models.CharField(max_length=50)
+    lat = models.DecimalField(max_digits=22,
+    decimal_places=16)
+    lon = models.DecimalField(max_digits=22,
```
ZakkMan: Can you use the full name for lat and long?

api/cotrip_app/models.py
```diff
@@ -6,3 +6,17 @@ class Topic(models.Model):
...
+    decimal_places=16)
+    city_image = models.CharField(max_length=200)
```
ZakkMan: We'll want to figure out image uploads. Can you see if that's something we can implement here?

api/cotrip_app/models.py
```diff
@@ -6,3 +6,17 @@ class Topic(models.Model):
 
     def __str__(self):
         return self.title
+
+class Location(models.Model):
```
ZakkMan: How does the Location model relate to the City model? Are they the same?

api/cotrip_app/views.py
```diff
@@ -1,3 +1,16 @@
...
 # Create your views here.
+
+class LocationList(generics.ListApiView):
+    queryset = Location.objects.all()
+    serializer_class = LocationSerializer
+
+class LocationDetail(generics.RetrieveUpdateDestroyAPIView):
+    queryset = Location.objects.all()
+    serializer_class = BookSerializer
+    permission_classes = (permissions.IsAuthenticated)
```
ZakkMan: We probably don't want people to be able to update, create, or delete these other than through the Django admin

api/cotrip_app/models.py
zfriendly: I don't think there is a City model, so figured Location made more sense. That way for the city property we could put in something like "The Yucatan" or something like that.

api/cotrip_app/models.py
zfriendly: Working on it!

api/cotrip_app/models.py
zfriendly: Changed.

api/cotrip_app/views.py
zfriendly: True.

zfriendly:
Made a few changes, removing CRUD. Turns out we're using Pillow for image uploads in the profile model (from SEI 33) so I just used the same.

PR #447, opened 2020-03-30
changed structure for pull request template

caioingber:


caioingber:
@ZakkMan @Underwaterr I was able to set up issue templates -> merging so PR template should show by default

PR #462, opened 2020-03-31
Component: Person Card-Online Green Dot

hisstori:
Issue #333 

Created the online indicator component to be mounted when a user is logged into the app currently.

Feature is complete. (Edit: I feel like I was supposed to add this to the PersonCard as a whole. Will update and tag once this is completed.)

No known bugs or strange behavior with this feature.

I would like to leave my own feedback here instead. I think it would be beneficial to the users if we created another component for when a user is logged in but has been inactive for 10 minutes or longer. Essentially an "IDLE" dot.

- Screenshot(s):
Created feature below -


hisstori:
Modified logic to render a green dot when a user is online.

client/src/components/PersonCard/PersonCard.js
```diff
@@ -3,8 +3,26 @@ import { Link } from "react-router-dom";
 import "./PersonCard.css";
 import ProfilePicture from "../ProfilePicture/ProfilePicture";
 import image from "../../assets/images/profile-picture-1.png";
+
 // Function based React Component
 const PersonCard = props => {
+  if (online) {
```
ZakkMan, line 9: Rather than a massive conditional like this where most of this is repeated, can we do a single conditional to show or hide the dot? Use the `&&` trick!

client/src/components/PersonCard/PersonCard.js
```diff
@@ -29,7 +47,12 @@ PersonCard.defaultProps = {
     { url: "#", interests: "budgeting" },
     { url: "#", interests: "travel" },
     { url: "#", interests: "outdoors" }
-  ]
+  ],
+  dot: <div className="PersonCard__online"></div>,
 };
 
+// Once user logs in online should be set to true and display green dot.
```
ZakkMan, line 54: Do we need these comments?

client/src/components/PersonCard/PersonCard.js
```diff
@@ -29,7 +47,12 @@ PersonCard.defaultProps = {
     { url: "#", interests: "budgeting" },
     { url: "#", interests: "travel" },
     { url: "#", interests: "outdoors" }
-  ]
+  ],
+  dot: <div className="PersonCard__online"></div>,
```
ZakkMan, line 51: What's this doing? Shouldn't the prop be something like `status` or `online`? Like a boolean value that controls whether or not the dot is visible

PR #465, opened 2020-03-31
updated user name to show last name initial

Mdusseau:
#419

updated UI

yes 

none
no

- Screenshot(s):


client/src/pages/HomePage/HomePage.js
```diff
@@ -79,6 +79,8 @@ const handleClick = e => {
 };
 
 const HomePage = props => {
+      console.log(localStorage.token);
```
ZakkMan, line 82: Can you remove any `console.log` statements?

client/src/pages/MemberProfilePage/MemberProfilePage.js
```diff
@@ -46,7 +46,8 @@ class MemberProfilePage extends Component {
   render() {
     let people = [
       {
-        name: "Suzie",
+        userFirstName: "Suzie",
+        userSurname:"Carmicheal",
```
ZakkMan, line 50: If this is for the user, then the properties don't need `user` in them (it's redundant)

client/src/pages/SplashPage/SplashPage.js
```diff
@@ -15,10 +16,21 @@ class SplashPage extends React.Component {
   toRegister() {
     this.props.history.push("/register");
   }
```
ZakkMan, line 18: Doesn't look like you're using this method. Can you delete it?

client/src/components/Slider/Slider.js
```diff
@@ -0,0 +1,54 @@
+import React, { Component } from "react";
```
ZakkMan, line 1: These files for the Slider component are part of a separate PR - can you remove them from this PR?

client/src/pages/MemberProfilePage/MemberProfilePage.js
```diff
@@ -46,14 +45,15 @@ class MemberProfilePage extends Component {
   render() {
     let people = [
       {
-        name: "Suzie",
+        FirstName: "Suzie",
+        Surname: "Washington",
```
ZakkMan, line 49: Why is the first character capitalized here? In JS, we use camelCase

client/src/pages/MemberProfilePage/MemberProfilePage.js
```diff
@@ -125,11 +125,10 @@ class MemberProfilePage extends Component {
                   facebook={people[0].facebook}
                   pinterest={people[0].pinterest}
                   type="default"
-                  name={people[0].name}
+                  name={people[0].FirstName + " " + people[0].Surname[0] + "."}
```
ZakkMan, line 128: Maybe this logic should be in the component, since we want it to be applied consistently.

client/src/pages/SplashPage/SplashPage.js
ZakkMan: Looks like you missed this one

client/src/pages/SplashPage/SplashPage.js
```diff
@@ -15,10 +16,21 @@ class SplashPage extends React.Component {
...
+
+  renderRedirect = () => {
+    if (localStorage.getItem("token") != undefined) {
+      return <Redirect to="/home" />;
+    } else{
+      console.log("Not logged in")
+    }
+  };
```
ZakkMan, line 26: This logic should probably be part of the parent component. If the user is logged in (and there's a token) then we want to just render the home page and not render the login page. here, you're rendering the login page, then checking to see if the user is logged in, and if they logged in only then redirecting them to the home page.

Issue #89, opened 2019-12-06
Component Molecular: Forum Post / Comment | Location Form group | Issue


nickmckenney:
<img width="626" alt="Screen Shot 2019-12-06 at 2 45 14 PM" src="https://user-images.githubusercontent.com/55772458/70351486-071f9680-1837-11ea-9da9-684fae188342.png">

--- Post {Component} ----

{Component: Profile Picture with User Name}
`Image`: file.png(Circle)
`User Name`: Name of User
{Component}

`Title`: "Post Title": String {Component}

`HashTags`: Text {Component}

`Date`: ex(Aug 19 6:00pm) {Component}

`PostText`: Text

`Likes/Comments` : Number and String {Component}

`Topic Pill`: String

`Like Button / Comment Button`: Button {Component}

--- Comment {component} ---

{Component: Profile Picture with User Name}
`Image`: file.png(Circle)
`User Name`: Name of User
{Component}

`Text`:String
`Date`: ex(Aug 19 6:00pm) {Component}

`Likes/Replies`: Number and String {Component}

CSS Information:
Fourm Post:
```
position: absolute;
left: 0%;
right: 0%;
top: 0%;
bottom: 47.48%;
background: #FAFAFA;
border: 1px solid rgba(37, 37, 37, 0.5);
box-sizing: border-box;
border-radius: 3px;
```
Fourm Like/Comment Button:
```
position: absolute;
left: 0%;
right: 0%;
top: 0%;
bottom: 0%;

background: #E5E5E5;
border: 1px solid rgba(37, 37, 37, 0.5);
box-sizing: border-box;
border-radius: 3px;
```
Fourm Comment:
```
position: absolute;
left: 0%;
right: 0%;
top: 62.58%;
bottom: 0%;
background: #C4C4C4;
border: 1px solid rgba(37, 37, 37, 0.5);
box-sizing: border-box;
border-radius: 3px;
```

shelbyvjacobs:
This looks like it may be the parent of #80 Post component and #85 Comment component. Do you agree, @cjohnston96?

PR #468, opened 2020-04-01
Component Molecular: Forum Post / Comment | Location Form group

reisrael4:

#89 

Created a parent container for the ForumPost and Comment components so that a post may be displayed with its corresponding comments

I think so, it looks like the mock and it's not throwing any errors.

None that I'm aware of.

Do I need to include some sort of loop and store the comments in an array in order to display multiple comments?

- Screenshot(s):

Screen Shot 2020-04-01 at 10 29 54 AM


client/src/components/Comment/Comment.css
```diff
@@ -3,14 +3,19 @@
...
   background-color: var(--gray-dark);
   font-size: 14px;
   font-family: var(--paragraph-font);
   padding: 1.2rem;
 }
 
+img{
+  height: 1.5rem;
+  padding: 0;
+}
```
ZakkMan, line 17: Can you namespace this according to the css naming convention? Otherwise these styles will clash with other styles on the `img` tag

client/src/components/ForumPostContainer/ForumPostContainer.css
```diff
@@ -0,0 +1,15 @@
...
+  align-items: center;
+  justify-content: center;
+  overflow: auto;
+}
+
+.Comment {
+  margin-top: 7px;
+  background: #c4c4c4;
+  border: 1px solid rgba(37, 37, 37, 0.5);
+  /* box-sizing: border-box; */
```
ZakkMan, line 13: Can you delete any commented out code?

client/src/components/ForumPostContainer/ForumPostContainer.css
```diff
@@ -0,0 +1,15 @@
...
+  flex-flow: column nowrap;
+  align-items: center;
+  justify-content: center;
+  overflow: auto;
+}
+
+.Comment {
+  margin-top: 7px;
+  background: #c4c4c4;
+  border: 1px solid rgba(37, 37, 37, 0.5);
```
ZakkMan, line 12: Are there CSS variables for these colors that you could use?

client/src/components/ForumPostContainer/ForumPostContainer.js
```diff
@@ -0,0 +1,44 @@
...
+            topics={props.forumPost.topics}
+            image={props.forumPost.image}
+            post={props.forumPost.post}
+          />
+        ) : (
+          <ForumPost />
+        )}
+      </div>
+      <div>
+        {props.comment ? (
```
ZakkMan, line 28: It's possible that there may be multiple comments on a post. So maybe `props.comment` should be `props.comments` (i.e. an array of comments) and then this becomes a loop with `.map()` or something?

client/src/components/Button/Button.css
```diff
@@ -69,23 +69,34 @@
...
   background-color: var(--gray-lite);
+  border: 1px solid var(--black-lite);
+  border-radius: 3px;
 }
 .Button--comment img,
 .Button--like img {
   margin-right: 0.5rem;
+  height: 1em;
+  padding: 0;
 }
```
ZakkMan, line 93: These two buttons look identical - can you combine them into one?

PR #475, opened 2020-04-01
filter made. functional is now class component

andrew-frazier6:

#449 

Original Component was a functional component. Couldn't create a proper search without making it a Class holding state. Once changed to a Class Component I also separated the data from inside of the BookATripImage component and then displayed the cards as what i filtered through in the data array. 

I believe to make this more dynamic and for future trips being added, it might be helpful to make a separate JSON file that includes all the trips and then when we add to the JSON file its automatically added to the Book a trip page.

Complete, checked functionality.

None

### DIFF
--- a/client/src/components/Banner/Banner.stories.js
+++ b/client/src/components/Banner/Banner.stories.js
@@ -198,3 +198,4 @@ storiesOf("Banners/Banner", module)
       />
     </Banner>
   ));
+// THIS IS THE SEARCH BANNER ON BOOK_A_TRIP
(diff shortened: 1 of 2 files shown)


client/src/components/Banner/Banner.stories.js
```diff
@@ -198,3 +198,4 @@ storiesOf("Banners/Banner", module)
       />
     </Banner>
   ));
+// THIS IS THE SEARCH BANNER ON BOOK_A_TRIP
```
ZakkMan, line 201: Do we need this comment?

ZakkMan:
@andrew-frazier6 didn't we decide to close this one until search was implemented in the backend?

Issue #469, opened 2020-04-01
Create Django App for 'Trip' Category Models and it's Models: Trip, Location, Activity


colleenobrien3:
To improve organization, create a new app to handle the trip-related models. Models suggestion:

Trip
     Location: Many to one relationship with Location model
     Date start:
     Date end:
     Attendees: Many to many relationship with profile model
     Activities: One to many relationship with Activities model

Location
     City
     State
     Country
     Groups: One to many rel. with groups model
     Members there: One to many rel. with profile model
     Trips: One to many relationship with trip model

Activity
     For kids: boolean
     For moms: boolean
     Location: Many to one rel. with location model
     Trip: Many to one rel. with trip model

PR #486, opened 2020-04-02
Set Up Basic Trip App

colleenobrien3:
- Related Issue (include '#'): #469 Create Django App for 'Trip' Category Models and it's Models: Trip, Location, Activity

- Description of changes made: set up trip app with Trip, Location , Activity models with preliminary serializers, urls, views, etc.

- Is the feature complete/bug resolved/etc..: models, serializers, urls, and views will likely need to be edited as we progress with the user stories. perhaps we should make more specific issues for this.

- Screenshot(s):


ZakkMan:
@colleenobrien3 @saralaffin I'm going to have @Underwaterr take this one to review :)

PR #487, opened 2020-04-02
fix for immutable error when trying to create a user in django interface

saralaffin:
- Related Issue (include '#'): none that I know of, but haven't looked very hard. I think the backend issues were all closed to start fresh.

- Description of changes made: Ran into an immutable data error and used .copy to get around it. Also added a conditional for the 'profile' so creating a user works in the Django dashboard.

Not sure how other people were getting the database to work, so this is just a suggestion.

### DIFF
--- a/api/accounts/views.py
+++ b/api/accounts/views.py
@@ -10,8 +10,12 @@ class RegisterAPI(generics.GenericAPIView):
     serializer_class = RegisterSerializer
 
     def post(self, request, *args, **kwargs):
-        profile_data = request.data.pop('profile')
-        serializer = self.get_serializer(data=request.data)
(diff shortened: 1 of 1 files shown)


saralaffin:
After setting up the backend and running `pipenv run python manage.py runserver`
I navigated to `http://127.0.0.1:8000/auth/register` seen here:

image

 and then after filling out the Django form and clicking POST I got:

image

Also, I'll change the variable name and add a comment.

saralaffin:
@Underwaterr I added a comment and changed the variable name.

Underwaterr:
Looks great! 💯

saralaffin:
Thanks @Underwaterr!

Issue #194, opened 2019-12-10
Responsive Member Profile Page


joshseven:

caioingber:
There is currently no implemented responsive design for the MemberProfile Page url path is /member-page.

PR #488, opened 2020-04-02
made page responsive

vmipfdr:
- Related Issue (include '#'): #194 

- Description of changes made:  I made the site responsive

- Is the feature complete/bug resolved/etc..:  Complete

- Any known bugs/strange behavior:  There are some responsive issues with some Components that I'll address in a separate issue

- Screenshot(s):


vmipfdr:
@ZakkMan I fixed a bunch of the stuff, and after removing CSS for the imported components, I filed a few issues.

client/src/pages/MemberProfilePage/MemberProfilePage.css
```diff
@@ -22,10 +22,9 @@
   text-align: left;
 }
 
-.MemberProfilePage .Groups__list {
+.MemberProfilePage .Groups {
```
ZakkMan, line 25: This is still overwriting the styling of another component. Anywhere you're doing this, it means you need to go and update that component, because you've found a case that it can't handle

client/src/pages/MemberProfilePage/MemberProfilePage.css
```diff
@@ -104,9 +103,66 @@
...
+  }
+
+  .MemberProfilePage .Connections {
+    /* border: 1px solid green; */
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    width: 90%;
+    margin: 5%;
+  }
```
ZakkMan, line 142: This is also overwriting styles in a different component

client/src/pages/MemberProfilePage/MemberProfilePage.js
```diff
@@ -30,7 +30,7 @@ class MemberProfilePage extends Component {
 
     // Default CSS class to apply to the Component
     this.state = {
-      classList: "MemberProfilePage"
+      classList: "MemberProfilePage",
     };
   }
```
ZakkMan, line 36: Can you delete the empty, unused methods?

client/src/pages/MemberProfilePage/MemberProfilePage.js
```diff
@@ -53,53 +53,53 @@ class MemberProfilePage extends Component {
...
-      { name: "Martha", bio: "sup yall" }
+      { name: "Martha", bio: "sup yall" },
     ];
     let testUsers = [
       {
         userId: 1,
         userFirstName: "Paula",
         userSurname: "Bannerman",
-        userPic: pic1
+        userPic: pic1,
```
ZakkMan, line 65: If they're user objects, then the properties don't need `user` in them

client/src/pages/MemberProfilePage/MemberProfilePage.css
```diff
@@ -104,9 +103,66 @@
...
+    margin: 1vw;
+    padding-bottom: 0vh;
+    justify-content: center;
+  }
+  .MemberProfilePage__left-top {
+    margin-left: 0vw;
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    /* border: 2px solid black; */
```
ZakkMan, line 119: Please delete any commented out code

client/src/pages/MemberProfilePage/MemberProfilePage.js
```diff
@@ -30,7 +30,7 @@ class MemberProfilePage extends Component {
 
     // Default CSS class to apply to the Component
     this.state = {
-      classList: "MemberProfilePage"
+      classList: "MemberProfilePage",
```
ZakkMan, line 33: Do you need this in state like this?

client/src/pages/MemberProfilePage/MemberProfilePage.css
vmipfdr: Sure.

client/src/pages/MemberProfilePage/MemberProfilePage.js
vmipfdr: Not sure why this was added.  The only change to the js file was adding one additional div + className.

Issue #324, opened 2019-12-16
component: Community page - "discover new groups"


dynasty2:
![image](https://user-images.githubusercontent.com/53455521/70919803-83f90000-1fef-11ea-93f5-4e60a09ea8a2.png)

![image](https://user-images.githubusercontent.com/53455521/70919816-8bb8a480-1fef-11ea-9ff0-5df7e65724e8.png)

Good morning.  We have a straggling component that belongs to the community groups page.  The functionality behind this component is not yet fully defined or a huge priority.

dynasty2:
![image](https://user-images.githubusercontent.com/53455521/70925152-beb36600-1ff8-11ea-9f66-f70df311835d.png)

Also this similar component that lives in the "my friends" page.

PR #415, opened 2020-03-30
Discover Groups component

ZakkMan:
Build out component described in #324. 

I had to import this thing to make it so the component would work.

PR #490, opened 2020-04-02
Component discover connections

hisstori:
- Related Issue (include '#'): Issue #324 

- Description of changes made: Created the "Discover New Connections" component.

- Is the feature complete/bug resolved/etc..: Feature completed.

- Any known bugs/strange behavior: No known bugs/issues.

- Screenshot(s):

### DIFF
--- a/client/src/components/PersonCardDiscover/PersonCardDiscover.css
+++ b/client/src/components/PersonCardDiscover/PersonCardDiscover.css
@@ -2,7 +2,7 @@
   display: flex;
   flex-direction: column;
   align-items: center;
-  background: var(--white);
+  /* background: var(--white); */
   width: 13.5rem;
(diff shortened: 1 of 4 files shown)


ZakkMan:
@hisstori I just looked through the existing cards and found this one (`Card, Discover New Groups`). That said, it looks like it needs a little tweaking to the font. Can you update that one? Sorry for the duplication here! I thought we would have caught all this.

hisstori:
@ZakkMan quick note that I also submitted a PR on that card as well. The issue has both the discover groups & connections. Both of these cards are supported on the figma in their respective areas. 

Here is the link to that PR, I could not tag you for review. 
https://github.com/CotripperPlatform/CoTrip/pull/415

Issue #472, opened 2020-04-01
Create Community App and It's Models


colleenobrien3:
Suggestions:

Group
     Location: many to one with location model
     Title:
     Members: many to many with profile model
     Events: one to many with events model
     Posts: one to many with post model
     Media: One to many with media model
     
Event
     Title
     Group
     Location
     Members

Topic
     Title
     Description
     Followers: Many to many with user
     Post: one to many with post
     Media: one to many with media
Hashtag
     Title
     Description
     Posts: one to many with post
     Media: many to many with media
     Followers: many to many with user
Media
     Image url
     Hashtags
     Topic

PR #492, opened 2020-04-03
Api community app

saralaffin:
- Related Issue (include '#'): #472 

- Description of changes made: created models for Group, Event, Topic, Hashtag, Media

- Is the feature complete/bug resolved/etc..: No. Left a bunch of comments in the code about other models. Also, locations have been commented out because those models aren't created yet. 

- Any known bugs/strange behavior: What I have now works in the Django admin page :) and media files get saved to my local repo! But pretty sure this won't work on the deployed Heroku:

> app’s dynos have an ephemeral filesystem. This means that all files that aren’t part of your application’s slug are lost whenever a dyno restarts or is replaced (this happens at least once daily).

from: https://devcenter.heroku.com/articles/s3


api/community/models.py
```diff
@@ -0,0 +1,60 @@
...
+    title = models.CharField(max_length=200)
+    members = models.ManyToManyField('accounts.CustomUser', blank=True)
+    group = models.ForeignKey(Group, on_delete=models.CASCADE, related_name='events', null=True)
+    # location = models.ForeignKey('trip.Location', on_delete=models.CASCADE, related_name='events', null=True)
+    
+    def __str__(self):
+        return self.title
+
+class Topic(models.Model):
+    # there is a Topic model already in cotrip_app, but I've added it here too because it's included in this issue request. Should be updated in accounts.Profile when location of Topic model is decided upon
```
ZakkMan, line 26: Maybe delete the other topic model?

api/community/models.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def __str__(self):
+        return self.title
+
+# Followers: Many to many with user
+#       taken care of in accounts.Profile
+# Post: one to many with post
+#       (should be) taken care of in the Post model
+
+class Hashtag(models.Model):
```
ZakkMan, line 38: How much of a difference is there between topic and hashtag? Should/could we merge these and maybe add a flag that distinguishes if it's a hashtag or a topic?

saralaffin: Not much of a difference for now, so deleted hashtag model and added boolean hashtag field to Topic model.

api/community/models.py
saralaffin: Deleted Topic from cotrip_app, but didn't run makemigration in fear of merge conflicts

ZakkMan:
@saralaffin Can I close this PR since the issue has been closed?

api/api/settings.py
```diff
@@ -44,8 +44,12 @@
     "accounts",
     "knox",
     'corsheaders',
+<<<<<<< HEAD
+    'community',
+=======
     'trip',
     'django_extensions'
+>>>>>>> develop
```
ZakkMan, line 52: Looks like you committed a merge! Oh no! Can you fix this?

api/community/models.py
```diff
@@ -0,0 +1,46 @@
...
+#         (should be) taken care of in the Post model
+
+    def __str__(self):
+        return self.title
+
+class Event(models.Model):
+    title = models.CharField(max_length=200)
+    members = models.ManyToManyField('accounts.CustomUser', blank=True)
+    group = models.ForeignKey(Group, on_delete=models.CASCADE, related_name='events', null=True)
+    # location = models.ForeignKey('trip.Location', on_delete=models.CASCADE, related_name='events', null=True)
```
ZakkMan, line 20: Hey, why is this commented out? Do we need it? Can we delete it?

api/community/models.py
```diff
@@ -0,0 +1,46 @@
+from django.db import models
+
+# Create your models here.
+
+class Group(models.Model):
+    # can't add Location in until model exists in trip app
+    # location = models.ForeignKey('trip.Location', on_delete=models.CASCADE, related_name='groups', null=True)
```
ZakkMan, line 7: Same here re deleting commented out code

PR #508, opened 2020-04-06
Backend Deployment Heroku File Changes (Deploy Complete)

natesanchez:
Description of changes made:
1.) Added a Procfile and a requirements.txt file for Heroku deployment purposes.
2.) Pipenv installed gunicorn for Heroku purposes again.
3.) Updated/added STATIC_ROOT in settings.py file because I kept getting a particular error on Heroku, and these changes fixed it.

Is the feature complete/bug resolved/etc..: Yes

Any known bugs/strange behavior: No

Is there specific feedback you would like on these changes:
Once this is merged with develop, continuous deployment/integration with be complete for both the front and backend. Will still need to set up backend PostgreSQL and add a correct production route.


natesanchez:
@Underwaterr You're right. Removed it 👍🏽

Issue #499, opened 2020-04-06
Refine Trip Model - checkout api-trip-app branch


colleenobrien3:
Preliminary model in api-trip-branch

Ensure fields match with ERD

PR #517, opened 2020-04-06
Trip model

reisrael4:
- Related Issue (include '#'): #499 

Updated the models to match the ERD, created appropriate serializers and URLs. 

We think, seems to match ERD

Cannot run migrations at this time. Seems to stem from an issue between the cotrip_app migrations and the accounts migrations.

- Screenshot(s):


Underwaterr:
I didn't have any issues with the migrations, was this addressed?

Issue #504, opened 2020-04-06
Events Model - checkout api-community-branch to see what's been done


colleenobrien3:
Match fields to ERD

Issue #503, opened 2020-04-06
Media Model - checkout api-community-branch to see what's been done


colleenobrien3:
Make sure fields align with ERD

Issue #501, opened 2020-04-06
Hashtag model - checkout api-community-app branch to see what's been done


colleenobrien3:
I think we should leave the hashtag and topic models separate, if we are having one allow for user creation and one not.

Issue #500, opened 2020-04-06
Refine Group Model - checkout api-community-app branch to see what's been done


colleenobrien3:
Checkover model to ensure it aligns with erd

PR #519, opened 2020-04-06
Updated Community App API

caioingber:
- Related Issue (include '#'): #500, #501, #502, #503, #504 

- Description of changes made: updated models in community app and configured views/urls/serializers

Confirmed that API paths are viewable/functioning in DRF

Additionally, we encountered migration issues with the trip and accounts apps which we also corrected by including a default value of timezone.now for the date/time fields.

- Is the feature complete/bug resolved/etc..: I believe so

- Any known bugs/strange behavior: The migrations mentioned above

- Screenshot(s):


api/community/views.py
```diff
@@ -1,3 +1,54 @@
...
-# Create your views here.
+class GroupList(generics.ListCreateAPIView):
+    queryset = Group.objects.all()
+    serializer_class = GroupSerializer
+    permissions_classes = (permissions.IsAuthenticated)
+
+class GroupDetail(generics.RetrieveUpdateDestroyAPIView):
+    queryset = Group.objects.all()
+    serializer_class = GroupSerializer
+    permissions_classes = (permissions.IsAuthenticated)
```
ZakkMan, line 14: Do we want to have full crud on all these resources through the API? I think probably just read for List and Detail. Right?

ZakkMan: Oh maybe not - can users create groups?

caioingber: @ZakkMan We were thinking that we could allow full CRUD for community related data models. Our understanding is that this content is facilitated and managed by users, whereas trip data would be delegated to admins.

PR #521, opened 2020-04-06
S3 connection setup

saralaffin:
- Related Issue (include '#'): (none, but loosely related to user profile model)

- Description of changes made: Followed this guide: https://devcenter.heroku.com/articles/s3-upload-python to implement an S3 connection. Set up heroku env variables, but haven't tested those yet. 

- Is the feature complete/bug resolved/etc..: Need to test on master branch (or separate deployed testing branch?) to see if heroku variables and connections are set up correctly.


client/src/pages/OnboardingPage/PersonalSettingsForm/PersonalSettingsForm.js
```diff
@@ -16,16 +16,59 @@ class PersonalSettingsForm extends Component {
...
+    xhr.onreadystatechange = () => {
+      if (xhr.readyState === 4) {
+        if (xhr.status === 200 || xhr.status === 204) {
+          this.setState({ profile: { ...this.state.profile, image: url } });
+        } else {
+          alert("Could not upload file.");
+        }
+      }
+    };
+    xhr.send(postData);
```
Underwaterr, line 65: Can you update this to use `fetch` instead of `XMLHttpRequest`? The former is a much cleaner, more modern way of handling requests. Alternatively, look into using [Axios](https://github.com/axios/axios).

api/accounts/views.py
```diff
@@ -75,3 +77,26 @@ class UserList(generics.ListAPIView):
     
 # add view to update a profile, it should only allow a profile to be updated when they have a token saying theyre logged in
 
+def sign_s3(request):
+  S3_BUCKET = os.environ.get('S3_BUCKET_NAME')
```
ZakkMan, line 81: This should go into the `settings.py` file and import it from there. Also, this will need to work locally

client/src/pages/OnboardingPage/PersonalSettingsForm/PersonalSettingsForm.js
```diff
@@ -16,16 +16,59 @@ class PersonalSettingsForm extends Component {
         dream_destination: "",
         first_name: this.props.setup.first_name,
         last_name: this.props.setup.last_name,
-        city_of_residence: this.props.filter.city
-      }
+        city_of_residence: this.props.filter.city,
+      },
     };
   }
+  getSignedRequest = (file) => {
```
ZakkMan, line 23: This method is something that would go into the service for users (can chat about that later)

client/src/pages/OnboardingPage/PersonalSettingsForm/PersonalSettingsForm.js
saralaffin: Thank you again for the help with axios! Should be switched over now.

client/src/pages/OnboardingPage/PersonalSettingsForm/PersonalSettingsForm.js
saralaffin: Will pause on that until that folder gets create by Nate

api/accounts/views.py
saralaffin: Done! separate cotripperdev bucket set up with separate development user that has restricted access only to the cotripperdev bucket. Will send out .env file, once pull request is merged.

PR #522, opened 2020-04-07
First draft of a util file/ services folder for connecting to correct API

saralaffin:
- Related Issue (include '#'): none that I know of

- Description of changes made: suggested util.py file that returns the correct base url depending on production or development environment. 

- Is the feature complete/bug resolved/etc..: almost complete if we go with a util file versus exporting a promise object. I included what I'm guessing the promise object would look like, but since Zakk mentioned it was really simple, and this wasn't (for me anyways) I thought I'd pause and ask for a review. 

- Any known bugs/strange behavior: Haven't tested the promise object, but the util file inserts the correct url. Ran into other issues with my backend models, so not fully tested. But I got `POST http://localhost:8000/auth/register 500 (Internal Server Error)` in my console, so I know it's trying to get to localhost:8000 instead of heroku :smiley_cat:


client/src/services/authorizeUser.js
```diff
@@ -0,0 +1,48 @@
...
+//           } else {
+//             this.setState({
+//               email: json.email,
+//               first_name: json.profile.first_name,
+//               image: json.profile.image
+//             });
+//           }
+//         });
+//     }
+//   }
```
ZakkMan, line 48: Can we delete this commented out code?

client/src/services/authorizeUser.js
```diff
@@ -0,0 +1,48 @@
+import getAPI from "./util";
+import App from "App";
+
+// var promiseObj = new Promise(executor);
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise
+let promiseObjAuthUser = new Promise(authorizeUser(authToken))
```
ZakkMan, line 6: This is cool! `fetch` automatically returns a promise, so I think you can just return `fetch`. It'll look something like this:

```js
// User.js
const authorizeUser = (authToken) => {
  return fetch(`${API_URL}/auth/user`, {
    headers: {
      Authorization: `Token ${localStorage.getItem("token")}`,
    },
  })
```

Then, in the login page:

```js
// pages/Login.js
import authroizeUser from ../services/User.js 

// ...

componentDidMount() {
	authorizeUser.then(json => {
		// whatever, probably set state
	})
}
```

This wont be exactly what you need, but it should be close

saralaffin:
ok, Nate is going to start fresh :leaves:

Issue #515, opened 2020-04-06
Reply Model within Forum App


colleenobrien3:
Not in ERD yet, maybe model it off of comment Model description in ERD

Issue #514, opened 2020-04-06
Comment Model within Forum App


colleenobrien3:
Fields listed in ERD

Issue #513, opened 2020-04-06
Post Model within Forum App


colleenobrien3:
Consult ERD for fields

PR #529, opened 2020-04-08
Forum App Models, Views, Serializers, Urls

colleenobrien3:
- Related Issue (include '#'): #513 , #514 , #515 

- Description of changes made: app models for reply, comment, and post, also serializers, views, and urls

- Is the feature complete/bug resolved/etc..: may need to edit relationships once we have seed data and once we have other apps merged

- Screenshot(s):


api/forum/models.py
```diff
@@ -1,3 +1,40 @@
 from django.db import models
+from accounts.models import Profile
 
-# Create your models here.
+
+class Post(models.Model):
```
ZakkMan, line 5: These models are all largely the same, is there a way to combine them into one?

For instance, could you just have a `Post` model with a field like `parent` that relates one Post instance to another Post instance?

api/forum/models.py
```diff
@@ -1,3 +1,16 @@
...
+    parent = models.ForeignKey('self', related_name='comments', on_delete=models.CASCADE, null=True, blank=True, default=1)
```
ZakkMan, line 6: With just the `ForeignKey` I think you'll only be able to have one 'comment'. I think you'll need a many-to-many.

api/forum/models.py
```diff
@@ -1,3 +1,16 @@
...
+    post_type = models.CharField(max_length=200, null=True, blank=True)
```
ZakkMan, line 7: What is this field for?

colleenobrien3:
I made it a manytomany and tested it in the django ui. A post can have multiple comments now. Also, I added the type field to differentiate comments from posts and replies. Each type of post has different styling, according to the figma, so I thought we needed some way of differentiating them. @ZakkMan

ZakkMan:
@colleenobrien3 can you update this with develop and then merge it?

Issue #523, opened 2020-04-08
Create Seed Data for Community App


caioingber:
Follow Data Models and create data in accordance with it.

https://docs.djangoproject.com/en/3.0/howto/initial-data/

PR #531, opened 2020-04-08
Fixtures community app

caioingber:
- Related Issue (include '#'): #523

- Description of changes made: Added fixtures for community app, updated serializers and added title return 

- Is the feature complete/bug resolved/etc..: For now, will add more depth for foreign keys and many to many relationships once models are up to date.

- Any known bugs/strange behavior: Not sure how to serialize the topics so that their titles are viewable, utilizing a `CharField()` didn't seem to work since they are in a list. 

- Is there specific feedback you would like on these changes: above

- Screenshot(s):


api/community/serializers.py
```diff
@@ -7,12 +7,15 @@ class Meta:
...
+    group = serializers.CharField()
+    location = serializers.CharField()
+
     class Meta:
         model = Event
         fields = '__all__'
 
 class TopicSerializer(serializers.ModelSerializer):
-    class Meta:
+    class Mœeta:
```
ZakkMan, line 18: wow what's this?!

api/community/serializers.py
```diff
@@ -22,6 +25,9 @@ class Meta:
         fields = '__all__'
 
 class MediaSerializer(serializers.ModelSerializer):
+    hashtag = serializers.CharField()
+    group = serializers.CharField()
+    topics = serializers.HyperlinkRelatedField(many=True, )
```
ZakkMan, line 30: This should be [HyperlinkedRelatedField](https://www.django-rest-framework.org/api-guide/relations/#hyperlinkedrelatedfield)

ZakkMan: This is throwing errors so until this is fixed I can't test out the fixtures

api/community/serializers.py
caioingber: lol

api/community/serializers.py
caioingber: removed serializer definition for topics for the time being

api/community/serializers.py
Underwaterr: 👽

Issue #525, opened 2020-04-08
Create Seed Data for Accounts App


caioingber:

caioingber:
https://docs.djangoproject.com/en/3.0/howto/initial-data/

PR #534, opened 2020-04-09
Fixtures accounts

saralaffin:
- Related Issue (include '#'): #525 

- Description of changes made: Created fixtures for 5 accounts

- Is the feature complete/bug resolved/etc..: Currently references the activities model which is not on develop yet. I tested the fixtures without activities and got the following results:
 
image

- Is there specific feedback you would like on these changes: I haven't changed any serializers and wasn't really sure what I should be looking for when I tested it... but here is a file! It also occurred to me that a password wasn't set anywhere, so maybe these user instances won't be able to be signed in?


saralaffin:
@ZakkMan I was able to load it without the locations field, but this PR includes the locations field assuming those fixtures will be added in soon. Checkout the loaddata command I used in the screenshot on my first comment. I was able to see the entries in psql, but the topics column didn't show up. Is that a serializer issue?

saralaffin:
Removed locations from profile data and renamed both init files (in accounts app and community app).
Before the change, I was following suit on the naming convention set up in the community app. I thought the file path would be descriptive enough.
Also, this model works assuming the community app data is seeded first :)

saralaffin:
checkout PR #538. Looks like some accounts fixtures have already been merged..

saralaffin:
Since the other file existed on develop, I transferred the seed data over into fixtures-accounts.json. Working without the activities referenced! Making sure to seed community app fixtures first.

Issue #535, opened 2020-04-09
Profile picture URL not saving to database correctly


saralaffin:
**Describe the bug**
When adding the URL from AWS, the database adds a localhost address in front of it. 

**To Reproduce**
1. Go to http://localhost:3000/register
1. Automatically routes to: http://localhost:3000/home
2. Click on components in the dev tools
3. Scroll down to the HomePage component
4. See error in the 'image' property field. 

**Expected behavior**
I expected to only see the AWS url, without the localhost in front.

**Screenshots**
![image](https://user-images.githubusercontent.com/58662371/78895449-e2356d00-7a3c-11ea-8f6d-4232e060eae7.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu Bionic Beaver
 - Browser: Chrome
 - Version: 80.0.3987.149

PR #536, opened 2020-04-09
Fix profile picture url bug

saralaffin:
- Related Issue (include '#'): #535 

- Description of changes made: edit accounts.profile model from image to char field. 

- Is the feature complete/bug resolved/etc..: resolved!


api/accounts/models.py
```diff
@@ -23,7 +23,7 @@ class Profile(models.Model):
         on_delete=models.CASCADE, primary_key=True
     )
     topics = models.ManyToManyField('community.Topic')
-    image = models.ImageField(default='default.jpg', upload_to='profile_pics')
+    image = models.CharField(max_length=150, blank=True)
```
Underwaterr, line 26: Where did "max_length=150" come from? What happens if the user uploads a file with (for example) a 255-letter filename?

saralaffin: @Underwaterr the file name is set by a folder name and a UUID :)

Issue #560, opened 2020-05-10
Add a route for Upcoming Trips


life2free:
**Is your feature request related to a problem? Please describe.**
Currently, On the "Book a Trip" page, Upcoming Trips are displayed on page. In backend, there is no route for this. I will add a route for Upcoming Trips.

**Describe the solution you'd like**
Search trips by time, the start_date is after current date and order by start_date

Issue #558, opened 2020-05-10
 Add a route for searching trip


life2free:
**Is your feature request related to a problem? Please describe.**
Yes, this feature is related to the issue #554. On the "Book a Trip" page, there is a search bar for searching trip, but in backend, there is no route for searching trips. I will add a route to search for trips.

**Describe the solution you'd like**
Add a route, when getting the list of trip, filtering the list by title.

PR #571, opened 2020-05-11
api feature branch

life2free:
related #558 and #560

1. edit trip list route "trips/", add the search feature, user can search trips by title and time range.
   1) search by title with fuzzy query, return the list which each trip's title contains the filter's title (ignore case)
   2) time range: user can get a list with selecting a time range(start_date or end_date)
     start_date: return the list which each trip's start_date >= the selected start_date
     end_date: return the list which each trip's end_date <= the selected end_date
2. add a new route "trips/upcoming" for Upcoming trips, user also can search trips by title and time range.


clarknoah:
@life2free can you please provide my an example of how I would do a query of the `/trips` route?

clarknoah:
This is great brother, closing out this issue!

Issue #605, opened 2020-05-18
Populating the database with locations


digitalfences:
**Is your feature request related to a problem? Please describe.**
Currently no way to add locations to the database besides initial seeding. This could be adequate, considering we do not want locations to be available to users to write to. 

**Describe the solution you'd like**
All locations that the app serves could be preloaded to the database with a seed file. I think this is easy enough that I can implement it easily myself, so I will do so.

**Describe alternatives you've considered**
An api call could be made to a maps API on the front end then passed to the database to create location data.

PR #606, opened 2020-05-18
Locations Added and Seeded

digitalfences:
https://github.com/CotripperPlatform/CoTrip/issues/605

This pull request maps every city in the US into the database as a Location so that when US based users enter the site they can immediately find their city of residence and associate it to their profile. This will allow us to populate drop down menus that search through location by name or state all over the site. It also ensures we do not run into issues with users creating, deleting, or changing the locations in our database.


clarknoah:
@digitalfences can you do a pull from develop and fix any merge conflicts? then I can check this out. It seems anytime I merge on thing, all the other branches need it to in order to merge correctly

Issue #608, opened 2020-05-19
Fix Onboarding Profile image upload


Daniel-Edminster:
Use Imgur API for uploading/compression. Lightweight and free forever.

PR #609, opened 2020-05-19
Replaced S3 with Imgur Upload for profile storage

Daniel-Edminster:
#608 

- Description of changes made
Added Imgur Profile upload

Uploads image to imgur, gets URL, and passes to state/db 

- Screenshot(s):


clarknoah:
@Daniel-Edminster I have some bad news :(, Per higher ups, we have been instructed to stick with AWS, sorry, we just got the word :/

PR #629, opened 2020-06-29
slider component

0xZakk:
- Related Issue (include '#'): #425 

- Screenshot(s):


Underwaterr (changes requested):
what is this you're just deleting the app

Underwaterr (changes requested):
you changed nothing
are you sure you are happy working here? we should talk to HR

Issue #424, opened 2020-03-30
As a user, I would like to see the homepage and not the login page when I go to the website base URL after logging in


Underwaterr:
Currently, if you visit http://localhost:3000 after logging in it will show you the login screen again. It should either redirect to or show the homepage at http://localhost:3000/home

PR #633, opened 2020-06-30
added redirect to splash page

bavjean:
- Related Issue (include '#'): #424 

Added redirect to splash page if user is logged in

yes

Unsure of best way to go about this one. Seems to work properly though.

- Screenshot(s):


client/src/App.js
```diff
@@ -112,7 +112,13 @@ class App extends Component {
...
       <div className="App">
         <main>
-          <Route path="/" exact component={SplashPage}></Route>
+          <Route
+          path="/"
+          exact
+          render={(routerProps) => (
+            <SplashPage loggedIn={this.state.logged_in} {...routerProps}/>
+          )}
+          ></Route>
```
0xZakk: Could we reverse the order here? Better to redirect people to the splash page if they are not logged in than what we're doing here. We did it this way originally so that we'd have something up while we were building the app. But it's more likely that you're going to be visiting this page and be logged in.

0xZakk (changes requested):
@bavjean This looks good! Great work.

See my comment. Basically, now that the app is more developed, we should switch the order here. By default go to the homepage and if someone is not signed in, then redirect them to the splash page.

Thank you!

0xZakk:
@bavjean Does this make sense?

bavjean:
Still unsure what the home page old user is for, I only added this new functionality for the main /home

0xZakk:
@bavjean I'm not sure what that's about either. Would it be easy to delete?

client/src/App.js
```diff
@@ -243,7 +254,7 @@ class App extends Component {
             path="/home"
             exact
             render={routerProps => (
-              <HomePage handle_logout={this.handleLogout} {...routerProps} {...this.state} />
+              <HomePage loggedIn={this.state.logged_in} handle_logout={this.handleLogout} {...routerProps} {...this.state} />
```
0xZakk: I think you can delete this whole route, since the new homepage route is just `/`

client/src/App.js
```diff
@@ -112,7 +112,18 @@ class App extends Component {
...
+            render={routerProps => (
+              <HomePage
+                loggedIn={this.state.logged_in}
+                handle_logout={this.handleLogout}
+                {...routerProps}
+                {...this.state}
+              />
+            )}
+          ></Route>
           <Route path="/coming_soon" exact component={ComingSoonPage}></Route>
```
0xZakk: Can you delete this route too?

client/src/App.js
```diff
@@ -112,7 +112,18 @@ class App extends Component {
     return (
       <div className="App">
         <main>
-          <Route path="/" exact component={SplashPage}></Route>
```
0xZakk, line 115: Is this the start of the sign in flow? Will users get redirected to the sign in flow if they're not logged in?

client/src/pages/HomePage/HomePage.js
```diff
@@ -79,144 +80,154 @@ const handleClick = e => {
...
-              {personCard1}
-              {personCard2}
-              {personCard3}
-              <a href="/home" className="HomePage__arrow lower-people" />
-            </div>
-          </span>
+  const loggedIn = props.loggedIn;
+  if (!loggedIn) {
+    return <SplashPage {...props}/>;
+  } else {
```
0xZakk: This logic may be better off in the route

client/src/App.js
0xZakk: In my comment below where I say "in the route", I mean in this function.

0xZakk: Oh and rather than render a different component, can you redirect the user to a different page?

0xZakk (changes requested):
@bavjean See my feedback - looks good!

bavjean:
I think I got it all sorted out. Let me know if there is anything else I should change.

bavjean:
There were quite a few other links that push you back to /home instead of /. May have missed some

client/src/App.js
```diff
@@ -109,11 +107,28 @@ class App extends Component {
...
+                {...this.state}
+              />
+            )}
+          ></Route> :
+          <Route
+            path="/"
+            exact
+            component={SplashPage}
+          ></Route>
+            }
```
0xZakk: This is cool! Can you just use [`<Redirect />`](https://reacttraining.com/react-router/web/api/Redirect) thought?

0xZakk (changes requested):
Looks great! It's really close.

Did you try and do a search all through the entire codebase in VS Code for `/home`?

client/src/App.js
bavjean: Oops forgot about that part

0xZakk (approved):
Looks good!

Issue #423, opened 2020-03-30
As a user I would like to see the profiles of other users


Underwaterr:

MilesHamilton:
![screencapture-localhost-3000-community-explore-people-1593542519673](https://user-images.githubusercontent.com/46730797/86164679-2423fb80-bae0-11ea-9b6f-0494e4ac0b96.png)

MilesHamilton:
![screencapture-localhost-3000-community-explore-people-1593542744185](https://user-images.githubusercontent.com/46730797/86164887-80871b00-bae0-11ea-9e28-2ba522bc1dd3.png)

made it possible for a user to see all other users from the data in the /profile path to the /community/explore-people path on the  front end with random images as placeholders for each field

PR #637, opened 2020-06-30
Community page

MilesHamilton:
#423 
users are able to view other users info based on the info stored in the /profile path

- Screenshot(s):
screencapture-localhost-3000-community-explore-people-1593542744185


0xZakk (changes requested):
@MilesHamilton You checked in the environment variables!! Delete those immediately!

PR #638, opened 2020-06-30
havihng a difficult time tracking this one down

know-mad:
#421 
Changed state in a bunch of different components.....tried tracking it back so far that i even lost track of what components i made changes to. At this point im lost in my own work

no
when registering a new account, the topics of interest selected do not get added to the users profile thus preventing from being able to render the topics in the users profile page
how far into the rabbit hole should i go before i ask for help?......at this point im like 6 components deep in trying to figure out where the form is submitting the topics of interest upon registering
- Screenshot(s):
now getting this error in the server when registering a new user
Screen Shot 2020-06-30 at 3 11 50 PM

### DIFF
--- a/client/src/components/Bio/Bio.js
+++ b/client/src/components/Bio/Bio.js
@@ -22,32 +22,33 @@ class Bio extends Component {
   }
 
   toggleEditMode = () => {
-      this.setState({ 
+      this.setState({
         editMode: !(this.state.editMode)
       })
-    
+
   }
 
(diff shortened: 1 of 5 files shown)


0xZakk (changes requested):
@know-mad I think you figured it out:

> when registering a new account, the topics of interest selected do not get added to the users profile thus preventing from being able to render the topics in the users profile page

Sounds like you need to create an issue and fix this first

PR #646, opened 2020-07-01
User account add social media links

secondaryfun:
- Related Issue (include '#'): https://github.com/CotripperPlatform/CoTrip/issues/422

On MemberProfilePage:  
** Social Media Icons now populate from the user profile.
** toggleEdit now provides text boxes to add social media usernames. Social media options are populated based on Social_media_type table.
** On submit, urls are added to user profile for each social media option. URLs are generated as follows:
```  
url: `https://www.${media.name.toLowerCase()}.com/${username}` 
```
** The url currently generated assumes a certain pattern for social media sites. Suggest adding a row to the Social_media_type table of:  "url_base"  .  This would be the base url which usernames are added to to reach a user's account.  By adding it to the Model, future admins can add or change the base URL, if future social media sites do not fit the current pattern.

*** Larger issue:  Links to an outside side take users away from the core functionality of the site and encourage users to connect elsewhere.  I would suggest modifying this functionality to enable direct posting of messages/likes/etc. from the CoTripper account to outside social media accounts, thereby keeping the user within the CoTripper system.

- Screenshot(s):
image
image

### DIFF
--- a/client/src/components/Bio/Bio.css
+++ b/client/src/components/Bio/Bio.css
@@ -11,6 +11,10 @@
   color: var(--purple);
 }
 
+.Bio .icons__div-edit {
+  flex-flow: column nowrap;
+}
+
 .icons__div div {
   color: var(--purple);
   margin-top: 1rem;
--- a/client/src/components/Bio/Bio.css
+++ b/client/src/components/Bio/Bio.css
@@ -20,6 +20,7 @@
 }
 
 .Bio .icons__div div, .Bio .icons__div-edit div {
+
   color: var(--purple);
 }
 
@@ -31,7 +32,6 @@
 
 .Bio .Bio__paragraph {
   font-family: var(--paragraph-font);
-
 }
 
 .Bio .Bio__span {
@@ -41,7 +41,7 @@
 .Bio .Bio__header {
   display: flex;
   justify-content: space-between;
-  color: var(--purple)
+  color: var(--purple);
 }
 
 .Bio .hashtag__container {
@@ -50,4 +50,8 @@
   font-family: var(--paragraph-font);
   max-width: 100%;
   flex-wrap: wrap;
+}
+
+.change__password__title {
+  color: #6a4c93;
 }
--- a/client/src/components/Bio/Bio.css
+++ b/client/src/components/Bio/Bio.css
@@ -15,9 +15,12 @@
   flex-flow: column nowrap;
 }
 
-.icons__div div {
-  color: var(--purple);
+.Bio .icons__div div, .icons__div-edit__media-entry-container {
   margin-top: 1rem;
+}
+
+.Bio .icons__div div, .Bio .icons__div-edit div {
+  color: var(--purple);
 }
 
 .Bio .Bio__subheader,
--- a/client/src/pages/MemberProfilePage/MemberProfilePage.js
+++ b/client/src/pages/MemberProfilePage/MemberProfilePage.js
@@ -72,7 +72,7 @@ class MemberProfilePage extends Component {
   }
 
   render() {
-    this.state.protectedProfileData ? console.log(this.state.protectedProfileData.social_media) : console.log('no profile data')
+    // this.state.protectedProfileData ? console.log(this.state.protectedProfileData.social_media) : console.log('no profile data')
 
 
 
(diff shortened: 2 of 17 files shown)


client/src/pages/MemberProfilePage/MemberProfilePage.js
```diff
@@ -26,212 +26,214 @@ import axios from 'axios';
...
+	}
+
+
+
+	// Runs after a component has been updated
+	componentDidUpdate() { }
+
+	// Runs right before a component is removed from the DOM
+	componentWillUnmount() { }
+
```
0xZakk: Can you delete these if you're not using them?

client/src/components/Bio/Bio.css
```diff
@@ -11,11 +11,18 @@
   color: var(--purple);
 }
 
-.icons__div div {
-  color: var(--purple);
+.Bio .icons__div-edit {
```
0xZakk: I don't see these classes in your JSX. Can you remove them?

Unless these are classes in another component. You should never overwrite stiles in a child component! That will make debugging the CSS for that component nearly impossible in the future.

client/src/pages/MemberProfilePage/MemberProfilePage.js
```diff
@@ -26,212 +26,214 @@ import axios from 'axios';
...
+											isCurrentUser={true}
+											onClick={console.log("Hello")}
+											social_media={this.state.protectedProfileData.social_media}
+										/>
+										: ''}
+
+
+
+								</div>
+								<div></div>
```
0xZakk: Sup with this?

0xZakk (changes requested):
@secondaryfun Good start here!

client/src/pages/MemberProfilePage/MemberProfilePage.js
secondaryfun: Will do.

client/src/components/Bio/Bio.css
secondaryfun: >  You should never overwrite stiles in a child component!
I'm just using what I was given...  

I use :
.icons__div-edit is used in line 221 of Bio.js.


.icons__div div  is not mine.  It is called elsewhere on the Bio.js page.  I'm guessing someone previous used it to fix a styling bug that crops up with the TextField.  I can refactor the Bio.js page if you want me to, but I think others in the cohort are using it right now as well.  I tried to keep my footprint light.

secondaryfun:
Bug identified:   The `.update()` method does not support writable nested fields by default.

Options:
```
		// Step 1:  pull Profile Social Media (PSA) entries that match userid. 
		// Step 2:  Get pk of any existing PSA entries that have a matching Social Media Type (SMT).
		// Step 3:  Run PATCH on each match.
		// Step 4:  Run CREATE on each non match.
		// Step 5:  Profit.

		// Alternative:
		// Eliminate Profile Social Media table. Put user entries inside Profile.social_media list.
		// Step 1: Run PATCH once and update all profile data.
```

secondaryfun:
Forgive the misunderstanding re: lists in tables.  I found a serializer library for Django's Rest framework that has a nested table write option.  The ticket should be fully functional now.

0xZakk (changes requested):
@secondaryfun This looks good! Great work

Unfortunately I'm getting an error when I try to load the member-page:

```
Uncaught TypeError: Cannot read property 'name' of null
```

0xZakk:
@Underwaterr Do you mind testing this out? I'm sometimes getting an error when I sign in and load the user profile page

0xZakk:
@secondaryfun Where did we net out on this one? I think you said you were able to reproduce the error on this branch and on develop?

secondaryfun:
The error you described is unrelated to this issue.  See https://github.com/CotripperPlatform/CoTrip/pull/658 for the fix.  @ZakkMan

secondaryfun:
@ZakkMan This one should be ready to review. The error of the page loading was related to an issue on the app.js page, which I addressed in a separate issue.

0xZakk:
#658 has to be merged before this one can be

0xZakk (changes requested):
@secondaryfun #658 is merged to develop. Can you update this branch with develop and fix any merge conflicts and test all this out once more for me?

secondaryfun:
The last push required a --force. (git pull went smoothly, but the push still had errors)  No errors since the push.

0xZakk (changes requested):
Merge into your other branch, fix that up, and close this one out

PR #657, opened 2020-07-08
users can now add and drop connections with other users

ccorkery97:
Issue #457 
Changes were done in Directory page where a user can now see their connections, see other users they are not connected to, add those users to their connection list, and drop users from their connection list. 
the feature is complete
no known bugs
some feedback on the design of the directory page would be useful because I wasn't sure where to show the users connections. 
- Screenshot(s):


0xZakk (changes requested):
Are you missing some code here?

api/.vscode/settings.json
```diff
@@ -1,3 +1,2 @@
 {
-    "python.pythonPath": "/Users/bmillsmc/.local/share/virtualenvs/api-NkaCWkFK/bin/python"
```
0xZakk, line 2: Can you remove this file?

api/accounts/views.py
```diff
@@ -64,11 +64,10 @@ class ProfileList(generics.ListAPIView):
 
 
 class ProfileDetail(generics.RetrieveUpdateDestroyAPIView):
+    queryset = Profile.objects.all()
```
0xZakk, line 67: What was the thinking behind changing this? The detail page should show the signed in users profile (which is what the deleted code did). Does this work the same way?

client/src/App.js
```diff
@@ -98,7 +99,9 @@ class App extends Component {
                 first_name: json.profile.first_name,
                 image: json.profile.image,
                 userid: json.id,
-                profileLoaded: true
+                profileLoaded: true,
+                connections: json.profile.connections,
```
0xZakk: Can you move this ajax request into a service? All ajax requests should be an a service in the `services/` directory

client/src/pages/DirectoryPage/DirectoryPeople.js
```diff
@@ -16,173 +16,324 @@ import image4 from "../../assets/images/profile-picture-4.png";
...
+
+  componentDidMount() {
+    this.getFriends()
+    friends = this.state.friends
+  }
+
+
+  
+  
+  handleClick = e => {
```
0xZakk: Your formatting here is really weird. Can you make sure you have prettier set up with VS Code with auto format on paste and on save?

client/src/pages/DirectoryPage/DirectoryPeople.js
```diff
@@ -16,173 +16,324 @@ import image4 from "../../assets/images/profile-picture-4.png";
...
+      headers: {
+        Authorization: `Token ${localStorage.getItem("token")}`
+      }
+    })
+      .then(res => res.json())
+      .then(json => {
+        this.makeFriends(json)
+      });
+  }
+
```
0xZakk: Should be in a service to

client/src/pages/DirectoryPage/DirectoryPeople.js
```diff
@@ -16,173 +16,324 @@ import image4 from "../../assets/images/profile-picture-4.png";
...
+  friendCards = (arr) => {
+    let length  = friends.length
+    for (let i = 0; i < length; i++) {
+      //change to get the ith in the list and that objects name and image
+      arr.push(
+        <div className="CommunityPage__momCard-single">
+        <PersonCard image={this.state.friends[i].image} name={this.state.friends[i].first_name} location="Washington D.C." />
+        <button className="FriendButton" onClick={this.removeFriend}>Remove Friend</button>
         </div>
```
0xZakk: All our JSX should be inside the `render` method. This works but it's generally a bad practice

client/src/pages/DirectoryPage/DirectoryPeople.js
```diff
@@ -16,173 +16,324 @@ import image4 from "../../assets/images/profile-picture-4.png";
...
+    this.friendCards(cards)
+    return cards
+  }
+
+  friendCards = (arr) => {
+    let length  = friends.length
+    for (let i = 0; i < length; i++) {
+      //change to get the ith in the list and that objects name and image
+      arr.push(
+        <div className="CommunityPage__momCard-single">
```
0xZakk, line 93: Can you use the CSS naming convention from he readme?

client/src/pages/DirectoryPage/DirectoryPeople.js
```diff
@@ -16,173 +16,324 @@ import image4 from "../../assets/images/profile-picture-4.png";
...
+  
+  showFriends = () => {
+    let cards = []
+    this.friendCards(cards)
+    return cards
+  }
+
+  friendCards = (arr) => {
+    let length  = friends.length
+    for (let i = 0; i < length; i++) {
```
0xZakk, line 90: Seems like you probably need an array method or something here

client/src/pages/DirectoryPage/DirectoryPeople.js
```diff
@@ -16,173 +16,324 @@ import image4 from "../../assets/images/profile-picture-4.png";
...
+      //same as the friends function
+      arr.push(
+        <div className="CommunityPage__momCard-single">
+        <PersonCard image={this.state.others[i].image} name={this.state.others[i].first_name} location="Washington D.C." />
+        <button className="FriendButton" onClick={this.sendRequest}>Add Friend</button>
+        </div>
+      )
+    }
+  }
```
0xZakk: Same here re array method and putting this in the `render` method

client/src/pages/DirectoryPage/DirectoryPeople.js
```diff
@@ -16,173 +16,324 @@ import image4 from "../../assets/images/profile-picture-4.png";
...
+  removeFriend = (e) => {
+    let nonFriend = e.target.parentElement.children[0].children[1].textContent
+    let connectionsList = this.props.connections
+    for (let i = 0; i < this.state.friends.length; i++) {
+      if (this.state.friends[i].first_name === nonFriend) {
+        connectionsList.splice(i, 1)
+      }
+    }
+    let data = { connections: connectionsList }
+    fetch(`http://127.0.0.1:8000/profile/${this.props.userid}`, {
```
0xZakk: Can you put this in a service and use the `base_url` constant? Otherwise, this will break in production

client/src/pages/DirectoryPage/DirectoryPeople.js
```diff
@@ -16,173 +16,324 @@ import image4 from "../../assets/images/profile-picture-4.png";
...
+      body: JSON.stringify(data)
+    })
+    .then(response => response.json())
+    .then(result => {
+      console.log('Success:', result);
+    })
+    .catch(error => {
+      console.error('Error:', error);
+    });
+    window.location.reload(true)
```
0xZakk: Nope. Nope. Nope.

Please use react router

client/src/pages/DirectoryPage/DirectoryPeople.js
```diff
@@ -16,173 +16,324 @@ import image4 from "../../assets/images/profile-picture-4.png";
...
+      method: 'PATCH',
+      headers: {
+        Authorization: `Token ${localStorage.getItem("token")}`,
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify(data)
+    })
+    .then(response => response.json())
+    .then(result => {
+      console.log('Success:', result);
```
0xZakk: You should delete any `console.log`s before making a PR. Also, is the user going to look in the console to see what's going on? Can you display something in the UI to show that this worked or threw an error?

client/src/pages/DirectoryPage/DirectoryPeople.js
```diff
@@ -16,173 +16,324 @@ import image4 from "../../assets/images/profile-picture-4.png";
...
+    })
+    .catch(error => {
+      console.error('Error:', error);
+    });
+  }
```
0xZakk: You should get the users's id that the currently signed in user wants to friend and send a request to the backend with just that. All you need is the user ID for the friend request and the currently signed in user

0xZakk (changes requested):
@ccorkery97 This looks pretty good and is a good start. See my comments below.

Generally speaking, I'd say this is more complicated than it needs to be. Most of the logic you have here should be happening on the backend and you shouldn't need this much logic.

For example, to create a friend request, the signed in user should click a button that sends a request to the backend with the ID for the user who is the target of the friend request (probably just in the URL parameters). The backend should take that ID and create the request between the user with that ID and the currently signed in User. You can have a status field on it (like `accepted`) which defaults to false. When that other user signs in, there should be a separate page where they can see all their friend requests (i.e. requests where `accepted` is false) and then submit a request to approve or reject the request.

api/accounts/views.py
ccorkery97: For some reason with the old code I could not see the detail page for every user but this change I don't believe is necessary, I only used it to see the details of other users. I can change it back.

client/src/pages/DirectoryPage/DirectoryPeople.js
ccorkery97: This is the naming convention already put into place. I wasn't sure if it would break anything on other pages if i changed it

ccorkery97:
I fixed some of the minor changes but wasn't sure what you wanted with the services, did you want me to make my own services for connections and add those functions in there? 

I did try putting the different cards into the render method but i couldn't get it to work and display each user in the array. 

With the request functionality I wasn't sure how to set up the backend so that the request array would have not only the id the request is from but also if it is accepted or not, so i just have and array of id's and when one is accepted it is removed from the array.

ccorkery97:
Before this PR can be merged: functions should be added to services, JSX should be put inside render, and the logic of the functionality should be done in such a way where requests have an ID field denoting from which user the request is and also a boolean field showing whether the request has been accepted or not.

Issue #656, opened 2020-07-08
Member Page fails to load after first login.


secondaryfun:
**Describe the bug**
After first logging in, the member page fails to load, leaving a blank screen. A soft refresh loads the page.

**To Reproduce**
0. Logout completely. Do a hard refresh and remove cookies.
1. Go to 'localhost:3000'
2. Login in.
3. From the home page, click on the profile image in top right corner.
4. See error:  White screen loads.

**Expected behavior**
Should see the member-page.

**Screenshots**
![image](https://user-images.githubusercontent.com/58237351/86923224-c0ab5680-c0fb-11ea-8098-90d5a39a358e.png)
![image](https://user-images.githubusercontent.com/58237351/86923255-cb65eb80-c0fb-11ea-867f-09482465bf0d.png)

**Desktop (please complete the following information):**
 - OS: Windows 10 
 - Browser: chrome 
 - Version  83.0.4103.116

PR #658, opened 2020-07-08
Moved getUserData(profile data fetch) from App to User and set to loa…

secondaryfun:
…d on login.

- Related Issue (include '#'):  https://github.com/CotripperPlatform/CoTrip/issues/656

** Set profile data fetch to occur following login.
** Moved fetch to services/user.js.

** Yes
** None I could find.

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -79,7 +79,7 @@ class App extends Component {
     this.getUserData = getUserData.bind(this);
   }
   componentDidMount() {
-    if (this.state.logged_in) this.getUserData()
+    if (this.state.logged_in) { this.getUserData() }
   }
 
 
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -36,7 +36,7 @@ import { fab } from "@fortawesome/free-brands-svg-icons";
 import { fas } from "@fortawesome/free-solid-svg-icons";
 import { far } from "@fortawesome/free-regular-svg-icons";
 import { BASE_URL } from "./services/constants";
-import { handleSignup, handleLogin, handleLogout } from "./services/User";
+import { handleSignup, handleLogin, handleLogout, getUserData } from "./services/User";
 
 library.add(
   fab,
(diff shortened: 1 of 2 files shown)


client/src/App.js
```diff
@@ -76,35 +76,14 @@ class App extends Component {
...
-                image: json.profile.image,
-                userid: json.id,
-                profileLoaded: true
-              },
-              this.logState
-            );
-          }
-        });
-    }
+    if (this.state.logged_in) this.getUserData()
```
0xZakk: Can you format this properly, with brackets?

client/src/services/User.js
```diff
@@ -47,7 +47,7 @@ export function handleLogin(data, history) {
         email: json.user.email,
         first_name: json.user.profile.first_name,
         image: json.user.profile.image
-      });
+      }, getUserData);
```
0xZakk: Are you chaining this method here? It would be better for the login endpoint to just return the data we need here rather than make a separate API call afterwards

client/src/services/User.js
```diff
@@ -56,6 +56,32 @@ export function handleLogin(data, history) {
...
 
+export function getUserData() {
+  fetch(`${BASE_URL}/auth/user`, {
+    headers: {
+      Authorization: `Token ${localStorage.getItem("token")}`
+    }
+  })
+    .then(res => res.json())
+    .then(json => {
+      // console.log(json);
```
0xZakk: Delete any commented out stuff throughout the PR

0xZakk (changes requested):
@secondaryfun See my feedback. Overall, looks good

client/src/App.js
secondaryfun: done

client/src/services/User.js
secondaryfun: done

client/src/services/User.js
secondaryfun: done

0xZakk (approved):

PR #673, opened 2020-07-13
Change user password

Las6103:
- Related Issue (include '#'): #425 

This branch contains Chris's social media adjustments as well as the password change feature.
The password feature is complete and fully functional.
NA
- Styling could be adjusted if necessary
- Screenshot(s):


client/src/components/Bio/Bio.css
```diff
@@ -44,3 +52,7 @@
   max-width: 100%;
   flex-wrap: wrap;
 }
+
+.password_header {
+  color: #6a4c93;
+}
```
0xZakk: Can you use the CSS naming convention?

0xZakk (changes requested):
Rather than have an alert pop down, can you display a message in UI? If the password is successfully changed, we should replace the form with a success message. If there's an error, then display that error with the form.

Also, can we do it without the email field? You can pull the email from the current user who is signed in. You want to check the JWT to verify auth anyway, rather than effectively have them re-sign in.

0xZakk (changes requested):
Add labels to inputs and then merge it

Las6103:
Merging the request. No conflicts.

PR #654, opened 2020-07-07
Community group page detail

secondaryfun:
- Related Issue (include '#'): https://github.com/CotripperPlatform/CoTrip/issues/580

- Description of changes made:
* Added dropdown selector for groups.
* Now populating the following from state:
** Title
** Post
** Comments
** Description
** Members

- Is the feature complete/bug resolved/etc..:

Features remaining to add as issues:
** View upcoming events for the group.
** View group media.
** Replies to comments
** Add comments via button
** Add likes to posts or comments

- Any known bugs/strange behavior:

* Currently, the group detail is set statically.  

- Is there specific feedback you would like on these changes:

- Screenshot(s):
image

PR #678, opened 2020-07-13
Community create group

bavjean:
- Related Issue (include '#'): #428 

Made a new community page for creating a new group

Still not entirely finished, need to redirect user after creation, and should probably add checker to make sure title and/or description are not empty.

Probably want to redirect to the view-group page but don't think that has been implemented yet.

Could use some css, and a better button to link to the page.(unsure best place to put it, so I just stuck it in the community join group page)

- Is there specific feedback you would like on these changes
I still need to add the things I mentioned above, mostly just wanted to get some feedback in case I need to change anything major.

I was originally using the form from the onboarding page(FilterSettingsForm) since it was exactly what I was trying to do, but realized the back end does not support interests within the group model. After considering changing the model, I decided it wouldn't make a whole lot of sense to include interests with this since the location and title should be enough to determine if the user wants to join. 

- Screenshot(s):


client/src/pages/CommunityPage/CommunityPage.js
```diff
@@ -151,7 +151,7 @@ const CommunityPage = props => {
         <a className="seeMore-Button">See More</a>
         <div>
           <header className="CommunityPage__header">
-            Groups in WASHINGTON, DC: <a href="./view-group"> View Sample Page </a>
+            Groups in WASHINGTON, DC: <a href="./view-group"> View Sample Page </a><a href="./create-group">Create Group</a>
```
0xZakk: You can remove the sample page link, since the group pages should now link to groups!

client/src/pages/CommunityPage/CommunityPageCreateGroup.js
```diff
@@ -0,0 +1,194 @@
...
+    axios({
+      method: 'POST',
+      url: `${BASE_URL}/groups`,
+      data: {
+        title: this.state.title,
+        description: this.state.description,
+        location: this.state.location
+      }
+    })
+  };
```
0xZakk: All ajax requests should go in a service in the `services/` directory

client/src/pages/CommunityPage/CommunityPageCreateGroup.js
```diff
@@ -0,0 +1,194 @@
...
+  componentDidMount() {
+    this.getUSStates();
+  }
+
+  getUSStates = () => {
+    axios.get(`${BASE_URL}/location/states`).then(res => {
+      console.log(res.data);
+      this.setState({ USStates: res.data });
+    });
+  };
```
0xZakk: This should also be in a service

client/src/pages/CommunityPage/CommunityPageCreateGroup.js
```diff
@@ -0,0 +1,194 @@
...
+      if (state.name.toLowerCase() === userSubmission) {
+        axios.get(`${BASE_URL}/location/bystate?state__code=${state.code}`).then(res => {
+          this.setState({
+            stateFound: true,
+            currentStateCities: res.data
+          });
+        });
+      }
+    });
+  };
```
0xZakk: This is going to be a lot of ajax requests - maybe we can load the cities once the state is selected, like we do on the account creation flow?

client/src/pages/CommunityPage/CommunityPageCreateGroup.js
```diff
@@ -0,0 +1,194 @@
...
+
+  setLocationID = event => {
+    let currentStateCities = this.state.currentStateCities;
+    // console.log(currentStateCities);
+
+    let userSubmission = event.target.textContent.toLowerCase();
+
+    // for(let i=0;i<currentStateCities.length)
```
0xZakk: Always delete commented out code!

client/src/pages/CommunityPage/CommunityPageCreateGroup.js
```diff
@@ -0,0 +1,194 @@
...
+            currentStateCities: res.data
+          });
+        });
+      }
+    });
+  };
+
+  setLocationID = event => {
+    let currentStateCities = this.state.currentStateCities;
+    // console.log(currentStateCities);
```
0xZakk: Delete commented out code and always remove any `console.log` statements

client/src/pages/CommunityPage/CommunityPageCreateGroup.js
```diff
@@ -0,0 +1,194 @@
...
+      return { found: true, location: cities[mid] };
+
+    // If element at mid is greater than x,
+    // search in the left half of mid
+    if (cities[mid].name.toLowerCase() > userSubmission)
+      return this.binarySearch(cities, userSubmission, start, mid - 1);
+    // If element at mid is smaller than x,
+    // search in the right half of mid
+    else return this.binarySearch(cities, userSubmission, mid + 1, end);
+  };
```
0xZakk, line 71: There has to be a better way to do this.

client/src/pages/CommunityPage/CommunityPageCreateGroup.js
```diff
@@ -0,0 +1,194 @@
...
+
+  render() {
+    return (
+      <div className="CommunityPage">
+        <Banner background={Banner__Community}>
+          <div className="community-page-header">
```
0xZakk: use the CSS naming convention in the documentation

0xZakk: *For all your CSS class names

0xZakk (changes requested):
@bavjean This is a great start! See my comments to your code, as well as my notes below.

This is what I see when I go to create a group:

Screen Shot 2020-07-14 at 10 29 43 AM

There are a couple of things you can do to really improve this form:

* Put a max-width on the container or wrap it somehow. Use your best judgement, but it probably shouldn't be much wider than like 500px.
* Put some space between the inputs
* Add labels to the inputs (in addition to the placeholder text)
* Give the form a title
* Make the button more visible. Buttons shouldn't be full width, instead have it be the width of the text with some padding and have it justified to the left.

This is a great resource on building forms: https://www.carbondesignsystem.com/patterns/forms-pattern

client/src/pages/CommunityPage/CommunityPageCreateGroup.js
bavjean: I used the account creation flow to do this part. I thought that with the conditional there it will wait to find a match for a valid state before loading the cities.

client/src/pages/CommunityPage/CommunityPageCreateGroup.js
bavjean: This is the only part I have not had time to mess with yet, this also came from the account creation flow.

0xZakk (changes requested):
Merge #654 and test everything out

Issue #688, opened 2020-07-14
A user should be able to search by name or by state and other profiles render


MilesHamilton:
A search feature that selectively renders profile pages when the user types in a state (or city) or name

PR #689, opened 2020-07-14
added changes from communityPage branch

MilesHamilton:
 old issue #427
 new issue #688 

added a search functionality that renders user profiles based on location or name

can only search for one field at a time and would have to empty the search bar before you can search again

feedback on search functionality, how to access a axios call from outside a functional component 

- Screenshot(s):


Underwaterr (approved):
Looks good!

